PR #51, opened 2019-02-20
various cleanup

dajohi:


network/client.go
```diff
@@ -61,13 +61,6 @@ func NewClient(conn net.Conn, endpoint *Endpoint, ip string) *Client {
 	}
 }
 
-// recordRequest logs a request as an id/method pair.
-// func (c *Client) recordRequest(id uint64, method string) {
```
dnldd: let's leave this commented out for now, there might be some stratum calls yet to be implemented that will use this.

network/client.go
```diff
@@ -76,19 +69,6 @@ func (c *Client) fetchRequest(id uint64) string {
 	return method
 }
 
-// deleteRequest removes the recorded request referenced by the provided id.
-// func (c *Client) deleteRequest(id uint64) {
```
dnldd: same here.

network/client.go
```diff
@@ -76,19 +69,6 @@ func (c *Client) fetchRequest(id uint64) string {
...
-// 	c.reqMtx.Lock()
-// 	delete(c.req, id)
-// 	c.reqMtx.Unlock()
-// }
-
-// nextID returns the next message id for the client.
-// func (c *Client) nextID() *uint64 {
```
dnldd: same here.

PR #68, opened 2019-05-04
harness: add logs 

jholdstock:
- harness: add logging
- harness: run dcrpool in repo root
    - this ensures that dcrpool is able to access local assets (eg. html templates)


jholdstock:
Copying comments from @dnldd on Matrix:

> the logs are good, can you add a line to the harness to copy the gui over to the harness folder instead?

> it should run in the harness folder

> we can add a config option to specify the path to load the gui on if copying over the gui folder into the harness doesn't work as expected

> think of the harness as a test sandbox, it's a mimic of how production would be

> we can't expect the gui to served from the project folder in production, it should be flexible/configurable

> with a configurable path the user can set things up as they see fit, thoughts?

PR #73, opened 2019-05-10
multi: gui cleanup.

dnldd:
This refactors gui related code and updates the config in preparation for more gui work.


gui/gui.go
```diff
@@ -0,0 +1,150 @@
...
+	httpTemplates, err := template.ParseFiles(templates...)
+	if err != nil {
+		return err
+	}
+
+	ui.templates = template.Must(httpTemplates, nil)
+	return nil
+}
+
+// run starts the user interface.
```
dajohi: Run

gui/index.go
```diff
@@ -0,0 +1,70 @@
...
+		PoolStats: poolStats,
+		HashRate:  poolHashRate,
+		HashSet:   hashSet,
+		SoloPool:  ui.cfg.SoloPool,
+		Admin:     false,
+	}
+
+	address := r.FormValue("address")
+
+	data.Address = address
```
dajohi: move after next if block

network/acceptedwork.go
```diff
@@ -175,6 +176,10 @@ func ListMinedWork(db *bolt.DB) ([]*AcceptedWork, error) {
 
 			if work.Confirmed {
 				minedWork = append(minedWork, &work)
+
+				if truncate && len(minedWork) == truncatedSize {
```
dajohi: check length to avoid a potential unnecessary allocation?

```go
if truncate && (len(minedWork) + len(work)) == truncatedSize {
    return nil
}
minedWork = append(minedWork, &work)
```

dnldd: `work` would still need appending to `minedWork` after checking `if truncate && (len(minedWork) + len(work)) == truncatedSize` before returning, no?

dajohi: you are correct.

dnldd:
Reverted changes regarding `backuppass`.

dnldd:
making additional modifications.

jholdstock:
additional modifications look good.

PR #75, opened 2019-05-13
network: fix accepted work confirmation.

### LINKED ISSUE #74: Failed to confirm accepted work.
Block #343018 (https://explorer.dcrdata.org/block/000000000000000007c092b0cd7cd5d83e8bafce5bf246d0f919e4326bd04408) was mined by the pool but was not confirmed by the next block as mined work. Finding the associated work for the block yielded an error instead: May 10 06:08:33 coordinator dcrpool[30780]: 2019-05-10 06:08:33.102 [ERR] NET: Failed to fetch accepted work: associated value for key '00053beb00000000000000001652cdf543a7d357fbe8c9e50452db562991337e72aa2f07' not found

Block #343018 was accepted at:
```
May 10 06:04:13 coordinator dcrpool[30780]: 2019-05-10 06:04:13.429 [INF] NET: Submited work hash at height (343018) is (000000000000000007c092b0cd7cd5d83e8bafce5bf246d0f919e4326bd04408)
May 10 06:04:13 coordinator dcrpool[30780]: 2019-05-10 06:04:13.429 [TRC] NET: pool target is: 118155209038499740525506916156177437716992928302701326536343361
May 10 06:04:13 coordinator dcrpool[30780]: 2019-05-10 06:04:13.429 [TRC] NET: hash target is: 190084330549982297427356466505204536126376274291409830920
May 10 06:04:13 coordinator dcrpool[30780]: 2019-05-10 06:04:13.432 [TRC] NET: Weighted share of (8776671248814965/281474976710656) for pool client (17725582/antminerdr5) claimed
```

The accepted work key that was not found (00053beb00000000000000001652cdf543a7d357fbe8c9e50452db562991337e72aa2f07) wen decomposed becomes:
```
~/Work/sandbox
❯ sandbox
height is: 343019  hash is: 00000000000000001652cdf543a7d357fbe8c9e50452db562991337e72aa2f07
```

**Conclusion**: An accepted work as not found for block #343019 at the time of checking, which resulted in the accepted work for block #343018 not getting marked correctly as a confirmed block. The process for confirming accepted work involves checking if a connected block's parent is an accepted work of the pool. If it is, it is marked as confirmed in order for payment processing to commence for the confirmed block. The solution for this would be to coordinate block connected notifications to checking for accepted work of a newly mined block to guarantee an accepted work is always created before the check is done. 

**Solution**: The block connected notification for a block already has the hash of the parent. There is no need to look for an accepted work of it first before fetching its parent. The previous block if mined by the pool is guaranteed to have its corresponding accepted work before the next block is connected.

dnldd:
This fixes the accepted work confirmation bug by fetching the parent accepted work of the connected
block instead of fetching the accepted work of the connected block first. This is being done because
it is possible for the accepted work of the connected block to be unavailable at the time its
corresponding block connected notification is received. It's also more efficient to directly fetch
the parent accepted work since that's ultimately the accepted work being confirmed.

Fixes #74.

### DIFF
--- a/network/acceptedwork.go
+++ b/network/acceptedwork.go
@@ -5,11 +5,11 @@
 package network
 
 import (
-	"bytes"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"strings"
+	"time"
 
 	bolt "github.com/coreos/bbolt"
 	"github.com/decred/dcrpool/database"
@@ -35,6 +35,7 @@ type AcceptedWork struct {
 	Height    uint32 `json:"height"`
 	MinedBy   string `json:"minedby"`
 	Miner     string `json:"miner"`
+	CreatedOn int64  `json:"createdon"`
 
 	// An accepted work becomes mined work once it is confirmed by an incoming
 	// work as the parent block it was built on.
@@ -58,6 +59,7 @@ func NewAcceptedWork(blockHash string, prevHash string, height uint32, minedBy s
 		Height:    height,
 		MinedBy:   minedBy,
 		Miner:     miner,
+		CreatedOn: time.Now().Unix(),
 	}
 }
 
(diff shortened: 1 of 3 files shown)


network/acceptedwork.go
```diff
@@ -35,6 +35,7 @@ type AcceptedWork struct {
 	Height    uint32 `json:"height"`
 	MinedBy   string `json:"minedby"`
 	Miner     string `json:"miner"`
+	CreatedOn uint64 `json:"createdon"`
```
dajohi: I would use int64 instead.

dajohi: it is also, set, but never read anywhere.  is that intended?

dnldd: Yes, the created on time coupled with the logs will be used in analyzing confirmations if there is a bug.

PR #78, opened 2019-05-19
gui: show connected miners

jholdstock:
### gui: Fix and simplify code.
This fixes some issues with the gui which were introduced
in the previous commit. Also:

- Combine methods in hub.go so pool stats can be fetched in one call.
- Remove truncate param from method to fetch mined work
- Remove minNano param from method to fetch payments
- Reverse slices in index.go so payments and mined blocks
are displayed in reverse order. (This is temporary)

### gui: show connected miners.
- index: show connected miners belonging to account
- move usage of util.HashString into the templates
- admin: show all connected miners
- Move AccountStats struct to gui package
- Add server side input validation to the Address form
- Combine functions FetchHash and FetchConnections so we only need to iterate
over connections once.

### DIFF
--- a/dividend/payment.go
+++ b/dividend/payment.go
@@ -555,10 +555,10 @@ func GeneratePaymentDetails(db *bolt.DB, poolFeeAddrs []dcrutil.Address, eligibl
 	return pmts, &targetAmt, nil
 }
 
-// FetchArchivedPaymentsForAccount fetches archived payments for the provided
-// account.
+// FetchArchivedPaymentsForAccount fetches the N most recent archived payments
+// for the provided account.
 // List is ordered, most recent comes first.
-func FetchArchivedPaymentsForAccount(db *bolt.DB, account string) ([]*Payment, error) {
+func FetchArchivedPaymentsForAccount(db *bolt.DB, account string, n int) ([]*Payment, error) {
 	pmts := make([]*Payment, 0)
 	err := db.View(func(tx *bolt.Tx) error {
 		pbkt := tx.Bucket(database.PoolBkt)
@@ -585,14 +585,15 @@ func FetchArchivedPaymentsForAccount(db *bolt.DB, account string) ([]*Payment, e
 				}
 
 				pmts = append(pmts, &payment)
+
+				if len(pmts) == n {
+					return nil
+				}
 			}
 		}
 
 		return nil
 	})
-	if err != nil {
-		return nil, err
-	}
 
-	return pmts, nil
+	return pmts, err
 }
--- a/dividend/payment.go
+++ b/dividend/payment.go
@@ -557,6 +557,7 @@ func GeneratePaymentDetails(db *bolt.DB, poolFeeAddrs []dcrutil.Address, eligibl
 
 // FetchArchivedPaymentsForAccount fetches archived payments for the provided
 // account.
+// List is ordered, most recent comes first.
 func FetchArchivedPaymentsForAccount(db *bolt.DB, account string) ([]*Payment, error) {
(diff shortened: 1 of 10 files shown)


gui/index.go
```diff
@@ -44,27 +61,54 @@ func (ui *GUI) GetIndex(w http.ResponseWriter, r *http.Request) {
...
-	data.AccountStats = resp
+	if len(work) == 0 && len(payments) == 0 {
+		log.Infof("Nothing found for address: %s", address)
+		data.Error = fmt.Sprintf("Nothing found for address: %s", address)
+		ui.renderTemplate(w, r, "index", data)
+		return
+	}
+
+	// // Reverse for display purposes. We want the most recent block/payments to be first.
+	reverseSlice(work)
```
dnldd: `ListMinedWork` already returns a reversed slice, it iterates and filters from the last entry to the first.

gui/index.go
```diff
@@ -44,27 +61,54 @@ func (ui *GUI) GetIndex(w http.ResponseWriter, r *http.Request) {
...
+	reverseSlice(payments)
```
dnldd: Update `FetchArchivedPaymentsForAccount` to iterate entries in reverse instead and remove `reverseSlice`.

Use: `for k, v := c.Last(); k != nil; k, v = c.First()`

jholdstock:
All review points have been addressed as we agreed on matrix.

- All data displayed on the GUI is now truncated to the most recent 10 entries. Temporary until we properly implement paging and can display all data efficiently
- Data is ordered during the db access using `Last` and `Prev`, rather than `First` and `Next`
- reverseSlice function removed
-

dividend/payment.go
```diff
@@ -571,28 +576,28 @@ func FetchArchivedPaymentsForAccount(db *bolt.DB, account []byte, minNano []byte
...
+
+				if len(pmts) == int(n) {
+					return nil
+				}
 			}
 		}
 
 		return nil
 	})
-	if err != nil {
```
dnldd: maintain this check, let's avoid heap allocation where we can.

network/acceptedwork.go
```diff
@@ -178,26 +182,27 @@ func ListMinedWork(db *bolt.DB, truncate bool) ([]*AcceptedWork, error) {
...
-				if truncate && len(minedWork) == truncatedSize {
+				if len(minedWork) == int(n) {
 					return nil
 				}
 			}
 		}
 
 		return nil
 	})
-	if err != nil {
```
dnldd: please maintain.

network/acceptedwork.go
```diff
@@ -219,16 +224,16 @@ func ListMinedWorkByAccount(db *bolt.DB, accountID string) ([]*AcceptedWork, err
...
 				minedWork = append(minedWork, &work)
+				if len(minedWork) == int(n) {
+					return nil
+				}
 			}
 		}
 
 		return nil
 	})
-	if err != nil {
```
dnldd: please maintain.

gui/index.go
```diff
@@ -44,27 +60,42 @@ func (ui *GUI) GetIndex(w http.ResponseWriter, r *http.Request) {
 
 	data.Address = address
 
-	minedWork, err := ui.hub.FetchMinedWork(true)
+	if len(address) != 35 {
```
dnldd: Sorry, missed this earlier. Please decode the address and assert it's derived from the active network set in the config. Refer to: https://github.com/decred/dcrpool/blob/master/config.go#L443-L457

jholdstock: added

jholdstock: added

PR #79, opened 2019-05-28
gui: move work quotas from admin page to homepage

jholdstock:
- validate payment scheme has been set correctly in config
- display the quota as a percentage rather than rational (ie. `20%` instead of `1/5`)
- if no payment quotas are found, return an empty map rather than an error
- do not return payment method whilst fetching work quotas. Take the payment scheme from config instead.

Screenshot from 2019-05-28 13-14-12

### DIFF
--- a/config.go
+++ b/config.go
@@ -399,6 +399,13 @@ func loadConfig() (*config, []string, error) {
 		return nil, nil, err
 	}
 
+	// Ensure a valid payment method is set.
+	if cfg.PaymentMethod != dividend.PPS && cfg.PaymentMethod != dividend.PPLNS {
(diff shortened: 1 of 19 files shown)


config.go
```diff
@@ -399,6 +399,13 @@ func loadConfig() (*config, []string, error) {
 		return nil, nil, err
 	}
 
+	// Ensure a valid payment method is set.
+	if cfg.PaymentMethod != dividend.PPS && cfg.PaymentMethod != dividend.PPLNS {
```
dnldd: This doesn't apply if the pool is solo pool mining. Also the payment method defaults to PPS if not set in the config, this check is not needed.

PR #80, opened 2019-05-28
config: change tls curve from P521 to P256

jholdstock:
This change stems from the fact that the dcrpool GUI was working in Firefox but not Chrome. It seems like FF has support for P521 but Chrome does not.

This PR changes to P256 which is supported by both. Further information on golang/go#19901


jrick:
I like this change.

Want to make P256 the default for dcrd and dcrwallet as well?

PR #83, opened 2019-06-06
multi: letsencrypt support.

dnldd:
This adds support for letsencrypt issued certs intended for use in production. Additional config options have been added to allow specifying the domain the cert should be issued for.

### DIFF
--- a/config.go
+++ b/config.go
@@ -48,6 +48,7 @@ const (
 	defaultSoloPool        = false
 	defaultGUIPort         = 8080
 	defaultGUIDir          = "gui"
+	defaultUseLEHTTPS      = false
 )
 
 var (
@@ -98,6 +99,8 @@ type config struct {
 	SoloPool        bool     `long:"solopool" ini-name:"solopool" description:"Solo pool mode. This disables payment processing when enabled."`
 	BackupPass      string   `long:"backuppass" ini-name:"backuppass" description:"The admin password, required for database backup."`
 	GUIDir          string   `long:"guidir" ini-name:"guidir" description:"The path to the directory containing the pool's user interface assets (templates, css etc.)"`
+	Domain          string   `long:"domain" ini-name:"domain" description:"The domain of the mining pool, required for TLS."`
+	UseLEHTTPS      bool     `long:"uselehttps" ini-name:"uselehttps" description:"This enables HTTPS using a Letsencrypt certificate. By default the pool uses a self-signed certificate for HTTPS."`
 	poolFeeAddrs    []dcrutil.Address
 	dcrdRPCCerts    []byte
 	net             *chaincfg.Params
@@ -260,6 +263,7 @@ func loadConfig() (*config, []string, error) {
 		SoloPool:        defaultSoloPool,
 		GUIPort:         defaultGUIPort,
 		GUIDir:          defaultGUIDir,
+		UseLEHTTPS:      defaultUseLEHTTPS,
 	}
 
 	// Service options which are only added on Windows.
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -21,7 +21,9 @@ require (
 	github.com/gorilla/sessions v1.1.3
 	github.com/jessevdk/go-flags v1.4.0
 	github.com/jrick/logrotate v1.0.0
+	github.com/mholt/certmagic v0.5.1
```
jrick: I would recommend using [autocert](https://godoc.org/golang.org/x/crypto/acme/autocert) instead.

config.go
```diff
@@ -474,8 +478,15 @@ func loadConfig() (*config, []string, error) {
...
 
+	// Ensure a domain is set if HTTPS via letsencrypt is preferred.
+	if cfg.UseLEHTTPS && cfg.Domain == "" {
+		return nil, nil, fmt.Errorf("a valid domain is required for HTTPS " +
+			"via letsencrypt")
+	}
+
 	// Generate the TLS cert and key if they not already exist.
-	if !fileExists(defaultTLSCertFile) || !fileExists(defaultTLSKeyFile) {
+	if cfg.UseLEHTTPS && (!fileExists(defaultTLSCertFile) ||
```
jholdstock: Is this correct? if UseLEHTTPS then generate a self-signed cert pair? Should this be `!UseLEHTTPS`?

gui/gui.go
```diff
@@ -179,22 +184,57 @@ func (ui *GUI) loadTemplates() error {
...
-}
 
-// Shutdown tears down the user interface.
-func (ui *GUI) Shutdown() {
-	ctx, cl := context.WithTimeout(ui.cfg.Ctx, time.Second*5)
-	defer cl()
-	if err := ui.server.Shutdown(ctx); err != nil {
-		log.Error(err)
-	}
+		if ui.cfg.UseLEHTTPS {
```
jholdstock: Could use `else` here rather than checking `UseLEHTTPS==true` and then checking `UseLEHTTPS==false`?

It would help make it really clear that we are only want to run one code block or the other, and never both.

gui/gui.go
```diff
@@ -179,22 +184,57 @@ func (ui *GUI) loadTemplates() error {
...
+			certCache := autocert.DirCache("certs")
+			ui.certMgr = &autocert.Manager{
+				Prompt:     autocert.AcceptTOS,
+				Cache:      certCache,
+				HostPolicy: autocert.HostWhitelist(ui.cfg.Domain),
+			}
+
+			// Redirect all regular http requests to their https endpoints.
+			go http.ListenAndServe(":80", ui.certMgr.HTTPHandler(nil))
```
jholdstock: Do we want any error handling on this? Eg. What if port 80 is already in use?

config.go
dnldd: you're right, good catch.

gui/gui.go
dnldd: For long code blocks if-else isn't ideal because "else" doesn't provide any detail on the condition being checked. You have to scroll back and infer. I think separating the code blocks, explicitly showing the conditions being checked reads better.

vctt94: The problem with else when having a verification with AND is because according to De Morgan law: 
 `~(P && Q) == (~P || ~Q)`

And it can end having complication and unexpected results. So I agree with @dnldd on this one.

jholdstock: Im not sure what you are referring to vctt94. These if conditions have no `OR`s or `AND`s.

Fair enough, lets leave it in, I just wanted to double check it wasnt an oversight

gui/gui.go
```diff
@@ -47,6 +53,7 @@ type GUI struct {
 	cookieStore *sessions.CookieStore
 	router      *mux.Router
 	server      *http.Server
+	certMgr     *autocert.Manager
```
jholdstock: There isnt any need for this to be added to the GUI struct. It can be declared and remain local inside the func `Run()`

dnldd: Noted, updating.

gui/gui.go
vctt94: sorry. I thought it was about https://github.com/decred/dcrpool/pull/83/files#diff-b4bda758a2aef091432646c354b4dc59R481

PR #84, opened 2019-06-07
miner: Make MaxProcs configurable

jholdstock:
Running the test harness on a laptop is threatening to burn my house down. CPU mining on all cores by default is not ideal when working on a laptop.

This makes the max procs configurable.

### DIFF
--- a/cmd/miner/main.go
+++ b/cmd/miner/main.go
@@ -13,8 +13,6 @@ import (
 )
 
 func main() {
-	// Use all processor cores.
-	runtime.GOMAXPROCS(runtime.NumCPU())
 
 	// Listen for interrupt signals.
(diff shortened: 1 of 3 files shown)


jrick:
A sleep could probably be added too to pace blocks better while consuming less energy.

cmd/miner/main.go
```diff
@@ -28,6 +26,8 @@ func main() {
 		return
 	}
 
+	runtime.GOMAXPROCS(cfg.MaxProcs)
```
dnldd: let's add a check to assert `MaxProcs` is less than or equal to `runtime.NumCPU()`. If it is more, default to `runtime.NumCPU()`.

jholdstock: Done

PR #87, opened 2019-06-07
gui: Fix minor issues.

jholdstock:
- Format unix time properly
- Include account ID when showing address details
- Link to admin page in footer
- Add spaces to table headers
- Links to dcrdata.
- Re-org columns of payments table.


jholdstock:
Rebased and ready for another look.

PR #95, opened 2019-06-22
multi: restructure packages.

dnldd:
This ultimately fixes a design issue that causes a cyclic import error with database upgrades.

This also adds the database upgrade missing from #90.


dnldd:
testing on dcrpool mainnet instance, looks good so far

gui/index.go
```diff
@@ -92,14 +92,14 @@ func (ui *GUI) GetIndex(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// Ensure address is on the active network.
+	// Ensure address is on the active pool.
```
jholdstock: This comment doesn't need changing

jholdstock:
Needs rebase

cmd/miner/config.go
```diff
@@ -139,11 +139,68 @@ func fileExists(name string) bool {
...
+}
+
 // newConfigParser returns a new command line flags parser.
 func newConfigParser(cfg *config, so *serviceOptions, options flags.Options) *flags.Parser {
 	parser := flags.NewParser(cfg, options)
 	if runtime.GOOS == "windows" {
-		parser.AddGroup("Service Options", "Service Options", so)
+		_, err := parser.AddGroup("Service Options", "Service Options", so)
+		if err != nil {
+			return nil
```
jholdstock: Should probably log the error here if we are not going to return it. This will help to prevent confusion later on.

config.go
```diff
@@ -221,11 +222,65 @@ func genCertPair(certFile, keyFile string) error {
 func newConfigParser(cfg *config, so *serviceOptions, options flags.Options) *flags.Parser {
 	parser := flags.NewParser(cfg, options)
 	if runtime.GOOS == "windows" {
-		parser.AddGroup("Service Options", "Service Options", so)
+		_, _ = parser.AddGroup("Service Options", "Service Options", so)
```
jholdstock: As with the other case, I think we should check for an error and log it here

config.go
```diff
@@ -221,11 +222,65 @@ func genCertPair(certFile, keyFile string) error {
...
 	}
 	return parser
 }
 
+// cleanAndExpandPath expands environment variables and leading ~ in the
+// passed path, cleans the result, and returns it.
+func cleanAndExpandPath(path string) string {
```
jholdstock: This method seems to be copy/pasted into `cmd/miner/config.go`. Can we remove this duplication?

dcrpool.go
```diff
@@ -139,14 +65,15 @@ func NewPool(cfg *config) (*Pool, error) {
 		MaxTxFeeReserve:   maxTxFeeReserve,
 		MaxGenTime:        new(big.Int).SetUint64(cfg.MaxGenTime),
 		PaymentMethod:     cfg.PaymentMethod,
+		DBFIle:            cfg.DBFile,
```
jholdstock: Typo `DBFIle` -> `DBFile`

config.go
dnldd: This is intentional, avoids importing function from dcrpool.

jholdstock:
looking good

PR #100, opened 2019-06-24
gui: display miner config info

### LINKED ISSUE #86: Replace VSP blurb with connection instructions
There is chunk of text on the home page which has been copy/pasted from VSP and needs updating

### LINKED ISSUE #99: Link to payment transactions on block explorer

jholdstock:
- Add miner configuration info to the home page (Closes #86)
- Add links to the block explorer for payments (Closes #99)
- Fix an issue where GUI was completely broken in solo pool mode. It was hitting an error attempting to retrieve work quotas which were not available. This section of the front page is no longer drawn when the pool is in solo mode.

Screenshot from 2019-06-24 21-57-05

### DIFF
--- a/gui/index.go
+++ b/gui/index.go
@@ -12,6 +12,7 @@ import (
 
 type indexData struct {
 	PoolStats        *network.PoolStats
+	PoolDomain       string
 	WorkQuotas       []network.Quota
 	PoolHashRate     *big.Rat
 	SoloPool         bool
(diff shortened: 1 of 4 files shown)


dnldd:
Nice, for the mining pool overview section I think you can remove the supported miners bit there, starting from 'The pool supports the following miners: ...'

That information can be inferred from the available ports to connect to

jholdstock:
done

PR #105, opened 2019-07-03
muti: fix client hash rate calculation.

### LINKED ISSUE #103: fix hashpower calculation
Noticed a the hash power calculation bug in testing where a D1 confirmed to have only 2 of 3  mining boards installed reported a hash rate of 48TH/s which is not possible with each board maxing out at 16 TH/s.

dnldd:
Resolves #103.

### DIFF
--- a/dcrpool.go
+++ b/dcrpool.go
@@ -6,6 +6,7 @@ package main
 
 import (
 	"context"
+	"math"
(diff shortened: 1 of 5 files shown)


dnldd:
Tested thoroughly on the dcrpool mainnet testing instance.

Issue #96, opened 2019-06-23
Automatic updates for the UI


jholdstock:
Can be achieved using web sockets.

Split out from #93

jholdstock:
@dnldd  please specify which fields you think should be automatically updated, and what kind of time period for refreshing the data?

dnldd:
the merged PR takes care of automatic update for the stats section. he reward percentages of participating needs automatic updates too (if the pool is in pool mining mode). This also goes for the blocks mined mined by pool list.

PR #107, opened 2019-07-16
gui: Update homepage pool stats every 10 seconds

jholdstock:
Adds a websocket endpoint to the webserver which is used to push updates to connected clients every 10 seconds. Currently updating Last Payment Height, Last Work Height and Pool Hash Rate. More to come...

I suggest reviewing commit by commit.

Part of #96


gui/gui.go
```diff
@@ -84,6 +84,30 @@ func (ui *GUI) route() {
...
+		if err != nil {
+			log.Error(err)
+			return
+		}
+
+		ui.SendUpdatedValues(poolStats)
+	}
+
+	// Use a ticker to push updates through the socket periodically
+	ticker := time.NewTicker(RefreshRate)
```
dnldd: let's have the goroutine exit cleanly via a context. Also lets limit the scope of the ticker to the goroutine it's used in:

```
go func(ctx context){
	ticker := time.NewTicker(RefreshRate)
	defer ticker.Stop()

	for {
		select{
		case <-ticker.C:
			socketUpdate()

		case ctx.Done():
			return
		}
	}
}()
```

gui/gui.go
```diff
@@ -84,6 +84,30 @@ func (ui *GUI) route() {
 	ui.router.HandleFunc("/admin", ui.PostAdmin).Methods("POST")
 	ui.router.HandleFunc("/backup", ui.PostBackup).Methods("POST")
 	ui.router.HandleFunc("/logout", ui.PostLogout).Methods("POST")
+
+	// Websocket endpoint allows the GUI to receive updated values
+	ui.router.HandleFunc("/ws", ui.RegisterWebSocket).Methods("GET")
+
+	socketUpdate := func() {
```
dnldd: `socketUpdate` isn't being reused, doesn't need to be a anonymous function as a result. Let's fetch the stats and call `SendUpdatedValues` in the goroutine on a tick instead.

gui/websocket.go
```diff
@@ -0,0 +1,56 @@
...
+	},
+}
+
+type Message struct {
+	PoolHashRate      string `json:"poolhashrate"`
+	LastWorkHeight    uint32 `json:"lastworkheight"`
+	LastPaymentHeight uint32 `json:"lastpaymentheight"`
+}
+
+const RefreshRate = 5 * time.Second
```
dnldd: `RefreshRate` should be unexported.

jholdstock:
Implemented suggested changes and rebased onto master. Ready for a second review.

dnldd:
Looks good.

PR #113, opened 2019-07-23
config: update tls file paths per the homedir.

dnldd:
This fixes a bug where the default tls file paths were being used even with a change of the homedir.


dnldd:
discussed this further with @dajohi, updating the config to alow specifying `--tlskey` and `--tlscert` to keep it consistent with dcrd and dcrwallet.

PR #120, opened 2019-07-26
multi: cache pool data.

### LINKED ISSUE #97: Cache UI info
We should not need to query all pool information every time the page is refreshed. This information can be cached.

Split out from #93

dnldd:
This adds caches for frequently queried data from the pool and cleans up the auto update processes for the gui. User facing calls have also been rate limited.

Fixes #97


pool/hub.go
```diff
@@ -1018,11 +1018,11 @@ type ClientInfo struct {
 	HashRate *big.Rat
 }
 
-// FetchClientInfo returns information about all connected clients
+// FetchClientInfo returns connection detials about all pool clients.
```
jholdstock: details*

gui/admin.go
```diff
@@ -36,30 +36,21 @@ func (ui *GUI) GetAdmin(w http.ResponseWriter, r *http.Request) {
...
 
-	if session == nil {
+	if session.Values["IsAdmin"] != true {
 		ui.renderTemplate(w, r, "login", pageData)
 		return
 	}
 
-	if session.Values["IsAdmin"] != true {
-		ui.renderTemplate(w, r, "login", pageData)
+	if !ui.limiter.WithinLimit(session.ID, pool.APIClient) {
```
jholdstock: This should be moved up, above the check for "IsAdmin".

gui/admin.go
```diff
@@ -122,21 +116,22 @@ func (ui *GUI) PostBackup(w http.ResponseWriter, r *http.Request) {
...
 
-	if session == nil {
+	if session.Values["IsAdmin"] != true {
 		http.Redirect(w, r, "/admin", http.StatusSeeOther)
 		return
 	}
 
-	if session.Values["IsAdmin"] != true {
-		http.Redirect(w, r, "/admin", http.StatusSeeOther)
+	if !ui.limiter.WithinLimit(session.ID, pool.APIClient) {
```
jholdstock: Move above "IsAdmin" check

PR #127, opened 2019-08-15
pool: fix client read goroutine termination.

dnldd:
This fixes an issue where the read goroutine of a client lingers due to the client sending invalid messages.


degeri:
Tested looks good.

PR #128, opened 2019-09-17
multi: avoid busy waiting cpuminer goroutines.

dnldd:
This fixes cases where goroutines of the cpu miner busy wait causing the miner to hangin the process. Some cleanup has also been done. The expected cpu miner hash rate has been lowered to speed up harness tests and  the introductory blog post for dcrpool has been removed.

The harness has been updated to specify an appdata path to avoid modifying existing simnet setup and the fix to [off-by one nonce handling issues](https://github.com/decred/dcrd/commit/adff6a0bac899ced670663f964d78e4306703dd8) discovered in dcrd have been ported over.


intro.md
```diff
@@ -1,24 +0,0 @@
-# Introducing dcrpool.
```
davecgh: Did you intend to delete this file?

cmd/miner/client.go
```diff
@@ -105,17 +104,19 @@ func (m *Miner) subscribe() error {
 	return nil
 }
 
+// sleep pauses goroutine execution for the provided duration.
+func (m *Miner) sleep(duration time.Duration) {
```
davecgh: It doesn't really matter much, but I don't see why this is needed.  Why not just call `time.Sleep` directly?

intro.md
dnldd: yes, it has served its purpose and is no longer needed.

cmd/miner/client.go
dnldd: updated.

harness.sh
```diff
@@ -224,6 +224,7 @@ fi
 
 if [ "${NETWORK}" = "simnet" ]; then
 tmux send-keys "dcrd -C ${DCRD_CONF} \
+--appdata=${NODES_ROOT}/master \
```
jholdstock: Should probably set the log dir as well?

dnldd: The default log directory is derived of the appdata set.

PR #136, opened 2019-10-16
multi: subscribe to work notification updates. 

dnldd:
**depends on #135**

This switches the pool from polling for work to getting work notifications via dcrd's `notifywork` rpc. Timestamp-rolled current work has also been implemented.


pool/hub.go
```diff
@@ -416,10 +419,13 @@ func NewHub(ctx context.Context, cancel context.CancelFunc, httpc *http.Client,
...
 		},
 
-		// TODO: Switch to OnWork notifications when dcrd PR #1410 is merged.
-		// OnWork: func(headerE string, target string) {
-		// 	h.processWork(headerE, target)
-		// },
+		OnWork: func(headerB []byte, target []byte, reason string) {
+			// Process new work generated as a result of a new parent or
+			// new votes.
+			if reason == NewParent || reason == NewVotes {
```
davecgh: What about the `newtxns` case?  The workers shouldn't be notified they need to switch templates *now*, but they should opportunistically be mining the latest work as they finish a cycle.

davecgh: This says resolved, but I don't see anything setting the work in response to `newtxns`?

PR #140, opened 2019-11-13
multi: add payment request processing.

dnldd:
This implements payment processing infrastructure to the pool and fixes an off-by-one bug regarding pruning eligible payments based on the minimum allowed amount.


pool/payment.go
```diff
@@ -312,16 +314,29 @@ func FetchPendingPaymentsAtHeight(db *bolt.DB, height uint32) ([]*Payment, error
...
-func FetchEligiblePaymentBundles(db *bolt.DB, height uint32, minPayment dcrutil.Amount) ([]*PaymentBundle, error) {
+func FetchEligiblePaymentBundles(db *bolt.DB, height uint32, minPayment dcrutil.Amount, paymentReqs map[string]struct{}) ([]*PaymentBundle, error) {
 	maturePayments, err := FetchMaturePendingPayments(db, height)
 	if err != nil {
 		return nil, err
 	}
 
 	bundles := generatePaymentBundles(maturePayments)
-	for idx := 0; idx < len(bundles); idx++ {
+	for idx := len(bundles) - 1; idx >= 0; idx-- {
```
jholdstock: Is there a reason this is iterating backwards? Might be nice to explain this unusual loop with a comment

dnldd: Yeah, iterating backwards is convenient here because the slice gets updated (via deletion) during iteration. When that happens the current index decrements (`i--`), iterating backwards already takes care of the decrement if there should be a deletion. It's ideal as result.

jholdstock: Clever. Definitely worth a comment to ensure people don't try and change this in future

PR #143, opened 2019-12-18
pool: decouple components.

dnldd:
This decouples components initially bundled with the pool hub in order to better test them and ultimately improve test coverage for the pool.


pool/paymentmgr.go
```diff
@@ -0,0 +1,538 @@
...
+// payPerLastNShares generates a payment bundle comprised of payments to all
+// participating accounts within the lastNPeriod of the pool.
+func (pm *PaymentMgr) payPerLastNShares(coinbase dcrutil.Amount, height uint32) error {
+	percentages, err := pm.PPLNSSharePercentages()
+	if err != nil {
+		return err
+	}
+	var estMaturity uint32
+	coinbaseMaturity := pm.cfg.ActiveNet.CoinbaseMaturity
+	if coinbaseMaturity == 0 {
```
jholdstock: Coinbase maturity is never set to 0 in any of the dcrd netparam files

pool/paymentmgr.go
```diff
@@ -0,0 +1,538 @@
...
+	lastPaymentCreatedOn := uint64(payments[len(payments)-1].CreatedOn)
+	pm.setLastPaymentCreatedOn(lastPaymentCreatedOn)
+	err = pm.cfg.DB.Update(func(tx *bolt.Tx) error {
+		// Update the last payment created on time and prune invalidated shares.
+		err := pm.persistLastPaymentCreatedOn(tx)
+		if err != nil {
+			return err
+		}
+		minNano := time.Now().Add(-(time.Second * time.Duration(pm.cfg.LastNPeriod))).UnixNano()
+		return pruneShares(tx, minNano)
```
jholdstock: Is there going to be an issue if `pruneShares` fails but `LastPaymentCreatedOn` time has already been updated?

pool/paymentmgr.go
```diff
@@ -0,0 +1,538 @@
...
+
+	case PPLNS:
+		return pm.payPerLastNShares(coinbase, height)
+
+	default:
+		return fmt.Errorf("unknown payment method provided %v", cfg.PaymentMethod)
+	}
+}
+
+// isPaymentRequested asserts if a payment request exists for the
```
jholdstock: `asserts` isn't really correct here, would be better to say `checks`

pool/paymentmgr.go
```diff
@@ -0,0 +1,538 @@
...
+	for accountID := range pm.paymentReqs {
+		delete(pm.paymentReqs, accountID)
+	}
+	pm.paymentReqsMtx.Unlock()
+
+	if len(eligiblePmts) == 0 {
+		return nil
+	}
+
+	rand.Seed(time.Now().UnixNano())
```
jholdstock: I believe the function `rand.Seed` can be executed once when the process starts, I dont see any need to execute it repeatedly.

pool/paymentmgr_test.go
```diff
@@ -0,0 +1,762 @@
...
+	minPayment, err := dcrutil.NewAmount(2.0)
+	if err != nil {
+		t.Fatalf("[NewAmount] unexpected error: %v", err)
+	}
+	maxTxFeeReserve, err := dcrutil.NewAmount(0.1)
+	if err != nil {
+		t.Fatalf("[NewAmount] unexpected error: %v", err)
+	}
+	activeNet := chaincfg.SimNetParams()
+	activeNet.CoinbaseMaturity = 0 // allow immediately mature coinbases, for testing.
```
jholdstock: I've just noticed this here after writing my other comment which states that CoinbaseMaturity is never 0.

Do we really want to be testing with this value if the production code never uses it? Seems like it could introduce bugs because we are testing code which is never run in production, and we have code which is always run in production but never tested.

dnldd:
Hmm, I mostly split the commits based on the introduced component or its associated tests. I'm curious to know what made reviewing each commit difficult.

dcrpool.go
```diff
@@ -121,33 +115,41 @@ func newPool(cfg *config) (*miningPool, error) {
 		MinerPorts:            minerPorts,
 		MaxConnectionsPerHost: cfg.MaxConnectionsPerHost,
 	}
-
-	p.hub, err = pool.NewHub(p.cancel, p.httpc, hcfg, p.limiter)
+	p.hub, err = pool.NewHub(p.cancel, hcfg)
+	if err != nil {
+		return nil, err
+	}
+	err = p.hub.Listen()
```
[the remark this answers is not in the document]
dnldd: You're right, `Connect()` should preceed `Listen()` here, rectifying.

gui/gui.go
```diff
@@ -29,23 +28,43 @@ import (
 	"github.com/decred/dcrpool/pool"
 )
 
-// Config represents configuration details for the pool user interface.
```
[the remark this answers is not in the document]
dnldd: Removed it in preference of commenting the fields.

pool/paymentmgr.go
dnldd: Yes, and it is never going to be zero in production. This just allows running tests with a mature coinbase once a block it is mined.

pool/paymentmgr.go
dnldd: `LastPaymentCreatedOn` and `pruneShares` are executed as a unit here, it's a single read-write transaction, if there is a failure any update made is rolled back.

pool/paymentmgr.go
dnldd: good catch, will call it in the constructor.

pool/paymentmgr_test.go
dnldd: noted, I'm going to take another look at it.

dnldd: good catch, removed it and updated the tests accordingly.

pool/hub.go
```diff
@@ -339,6 +339,14 @@ func (h *Hub) Listen() error {
 	return nil
 }
 
+// CloseListeners terminates listeners created by endpoints of the hub. This
+// should only be used in the pool's shutdown process the hub is not running.
+func (h *Hub) CloseListeners() {
```
jholdstock: This is being called if an error occurs during creating the GUI - nice. Does it also need to be called when the application is shutting down after starting successfully (eg. CTRL+C)?

pool/paymentmgr.go
```diff
@@ -415,11 +415,25 @@ func (pm *PaymentMgr) isPaymentRequested(accountID string) bool {
 	return ok
 }
 
-// addPaymentRequest creates a payment request from the provided account.
-func (pm *PaymentMgr) addPaymentRequest(accountID string) {
+// addPaymentRequest creates a payment request from the provided account
+// if not already requested.
+func (pm *PaymentMgr) addPaymentRequest(addr string) error {
+	_, err := dcrutil.DecodeAddress(addr, pm.cfg.ActiveNet)
```
jholdstock: There are three places where `Account()` is preceeded by a call to `DecodeAddress()`.

Perhaps `DecodeAddress` can be moved inside of `Account()`? It would remove duplication and ensure the a new caller of `Account` doesnt forget to call `DecodeAddress` first

pool/hub.go
dnldd: Nope, context cancellation will take care of closing the listeners and terminating all running processes if the pool starts successfully.

pool/paymentmgr.go
dnldd: Good catch, rectifying.

PR #147, opened 2020-01-09
multi: add fetchPaymentsForAccount.

dnldd:
This fixes a channel sync issue with endpoint tests and adds `fetchPaymentsForAccount` for listing both pending and archived payments for an account.

### DIFF
--- a/pool/payment_test.go
+++ b/pool/payment_test.go
@@ -158,8 +158,6 @@ func testAccountPayments(t *testing.T, db *bolt.DB) {
 		}
 	}
 
-	time.Sleep(time.Millisecond * 10)
-
 	// Create archived payments for account X.
 	abx := makePaymentBundle(xID, count, amt)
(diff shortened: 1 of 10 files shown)


pool/payment_test.go
```diff
@@ -144,3 +144,63 @@ func testArchivedPaymentsFiltering(t *testing.T, db *bolt.DB) {
...
+	// Create pending payments for account X.
+	pbx := makePaymentBundle(xID, count, amt)
+	for _, pmt := range pbx.Payments {
+		err := pmt.Create(db)
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	time.Sleep(time.Millisecond * 10)
```
jholdstock: I dont see why this sleep is necessary. As far as I can tell, everything above, including the DB call in `pmt.Create`, is synchronous

dnldd: you're right, don't need to sleep there.

PR #148, opened 2020-01-13
Remove hub references from GUI component

jholdstock:


.gitignore
```diff
@@ -6,3 +6,4 @@ dcrpool
 logs
 *.log
 *.conf
+/pool/testdb
```
dnldd: testdb is deleted after tests, not sure why it should be added here.

gui/gui.go
```diff
@@ -68,12 +69,31 @@ type Config struct {
 	// AddPaymentRequest creates a payment request from the provided account
 	// if not already requested.
 	AddPaymentRequest func(addr string) error
+	// FetchMinedWork returns the last ten mined blocks by the pool.
+	FetchMinedWork func() ([]*pool.AcceptedWork, error)
+	// FetchWorkQuotas returns the reward distribution to pool accounts
+	// based on work contributed per the peyment scheme used by the pool.
```
dnldd: s/peyment / payment

dnldd: s /peyment /payment

.gitignore
jholdstock: The testdb is indeed removed successfully, however I wanted to add this to .gitignore because I often run tests whilst reviewing my own changes, so the testdb shows up in my git diffs.

Issue #146, opened 2020-01-06
UI Update


dnldd:
Initial screens for dcrpool's ui update: https://xd.adobe.com/view/a814493c-193a-4dfc-5735-0e6d459a6924-de25/. Need to first take another look at the gui component and make improvements where needed before this gets implemented.

jholdstock:
Love the new designs. I will assign this to myself and start making the changes soon

dnldd:
Sure, let's get the updates discussed to the gui component done first. I've asked @buck54321 to take a look at the gui component and make suggestions on things we can improve since he played a key role in polishing dcrdata's UI/UX to what it is now.

PR #149, opened 2020-01-17
Implement new design on user facing pages.

jholdstock:
Part of #146 

Still to do:
- Pagination
- Identify pending payments


dnldd:
Initial work looks good, keep going. Will wait for the missing design bits before we start fine tuning it. :+1:

PR #151, opened 2020-02-17
multi: utilize time.Duration in config options.

dnldd:
This updates  `--lastnperiod` & `--maxgentime` config options to use the `time.Duration` type.


pool/paymentmgr.go
```diff
@@ -297,7 +297,7 @@ func (pm *PaymentMgr) PPSSharePercentages() (map[string]*big.Rat, error) {
 // accounts based on work performed measured by the PPLNS payment scheme.
 func (pm *PaymentMgr) PPLNSSharePercentages() (map[string]*big.Rat, error) {
 	now := time.Now()
-	min := now.Add(-(time.Second * time.Duration(pm.cfg.LastNPeriod)))
+	min := now.Add(-pm.cfg.LastNPeriod)
```
jholdstock: Could we use `.Sub()` instead of `.Add()` with a negative value here? Same below.

version.go
```diff
@@ -25,7 +25,7 @@ const (
 // versioning 2.0.0 spec (http://semver.org/).
 const (
 	appMajor uint = 1
-	appMinor uint = 1
+	appMinor uint = 2
```
jholdstock: I guess the reason for incrementing the version number is because of the change of config file format. We don't have a `1.1.0` tag yet, should we create one before merging this PR?

dnldd: Good catch, got ahead of myself there. Will revert.

pool/paymentmgr.go
dnldd: Will be maintaining using `Add` here because `Sub` returns the difference (duration) of two time values, `Add` returns a time updated with a duration. 

This is from comments on `Sub`: 
```
...
// To compute t-d for a duration d, use t.Add(-d).
```

pool/difficulty.go
```diff
@@ -43,12 +44,13 @@ type DifficultySet struct {
 }
 
 // NewDifficultySet generates difficulty data for all supported mining clients.
-func NewDifficultySet(net *chaincfg.Params, powLimit *big.Rat, maxGenTime *big.Int) (*DifficultySet, error) {
+func NewDifficultySet(net *chaincfg.Params, powLimit *big.Rat, maxGenTime time.Duration) (*DifficultySet, error) {
+	genTime := new(big.Int).SetUint64(uint64(maxGenTime))
```
dnldd: Should be using `maxGenTime.Seconds()` here.

pool/paymentmgr.go
dnldd: using `time.Sub` here was ideal https://github.com/decred/dcrpool/pull/151/files#diff-8335d701d364d65ba3e363261132d302R505

config.go
```diff
@@ -78,43 +78,43 @@ var runServiceCommand func(string) error
...
+	GUIDir                string        `long:"guidir" ini-name:"guidir" description:"The path to the directory containing the pool's user interface assets (templates, css etc.)"`
+	Domain                string        `long:"domain" ini-name:"domain" description:"The domain of the mining pool, required for TLS."`
+	UseLEHTTPS            bool          `long:"uselehttps" ini-name:"uselehttps" description:"This enables HTTPS using a Letsencrypt certificate. By default the pool uses a self-signed certificate for HTTPS."`
+	TLSCert               string        `long:"tlscert" ini-name:"tlscert" description:"Path to the TLS cert file."`
+	TLSKey                string        `long:"tlskey" ini-name:"tlskey" description:"Path to the TLS key file."`
+	Designation           string        `long:"designation" ini-name:"designation" description:"The designated codename for this pool. Customises the logo in the top toolbar."`
+	MaxConnectionsPerHost uint32        `long:"maxconnperhost" init-name:"maxconnperhost" description:"The maximum number of connections allowed per host."`
```
jholdstock: Looks like a regression here. Should be `ini-name`. Same on the following line

pool/hub.go
```diff
@@ -138,11 +139,14 @@ func NewHub(cancel context.CancelFunc, hcfg *HubConfig) (*Hub, error) {
...
 	h.blake256Pad = generateBlake256Pad()
 	powLimit := new(big.Rat).SetInt(h.cfg.ActiveNet.PowLimit)
-	maxGenTime := new(big.Int).SetUint64(h.cfg.MaxGenTime)
+	maxGenTime := h.cfg.MaxGenTime
 	if h.cfg.SoloPool {
 		maxGenTime = soloMaxGenTime
 	}
 
+	log.Infof("Maximum work submission generation time at "+
+		"pool difficulty is %s.", h.cfg.MaxGenTime)
```
jholdstock: I think this should use `maxGenTime` rather than `h.cfg.MaxGenTime`

PR #152, opened 2020-02-19
miner: fix miner shutdown panic.

dnldd:
This updates miner shutdown processes to avoid prematurely closing channels or blocking on a send or receive when shutting down, all processes fully terminate before the miner shuts down.

### DIFF
--- a/cmd/miner/cpuminer.go
+++ b/cmd/miner/cpuminer.go
@@ -187,6 +187,7 @@ func (m *CPUMiner) solve(ctx context.Context) {
 			case <-ctx.Done():
 				m.miner.wg.Done()
 				return
+
 			default:
 				continue
 			}
@@ -208,18 +209,19 @@ func (m *CPUMiner) solve(ctx context.Context) {
 				m.workData.extraNonce2, m.workData.nTime, m.workData.nonce)
 			m.workCh <- req
 
+			// Stall to prevent mining too quickly.
+			time.Sleep(time.Millisecond * 500)
+
 		case false:
 			select {
 			case <-ctx.Done():
 				m.miner.wg.Done()
 				return
+
 			default:
 				// Non-blocking receive fallthrough.
 			}
 		}
-
-		// Stall to prevent mining too quickly.
-		time.Sleep(time.Millisecond * 500)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


cmd/miner/cpuminer.go
```diff
@@ -89,7 +88,10 @@ func (m *CPUMiner) hashRateMonitor(ctx context.Context) {
 // This function will return early with false when conditions that trigger a
 // stale block such as a new block showing up or periodically when there are
 // new transactions and enough time has elapsed without finding a solution.
-func (m *CPUMiner) solveBlock(ctx context.Context, headerB []byte, target *big.Rat, ticker *time.Ticker) bool {
+func (m *CPUMiner) solveBlock(ctx context.Context, headerB []byte, target *big.Rat) bool {
+	ticker := time.NewTicker(333 * time.Millisecond)
```
jholdstock: This ticker seems to be unused (I think it was also unused before this PR)

cmd/miner/cpuminer.go
```diff
@@ -194,18 +198,28 @@ func (m *CPUMiner) solve(ctx context.Context) {
...
+			select {
+			case <-ctx.Done():
+				m.miner.wg.Done()
+				return
+			default:
+				// Non-blocking receive fallthrough.
+			}
 		}
+
+		// Stall to prevent mining too quickly.
```
jholdstock: This seems to have moved. Previously it was only executed when `solveBlock` returned true, now it is executed on every loop iteration. Is this correct?

cmd/miner/cpuminer.go
dnldd: The ticker is used here https://github.com/decred/dcrpool/blob/399537d35c0d81933a4dab53e68d182a388c0a5b/cmd/miner/cpuminer.go#L116, it was passed in as an argument before.

cmd/miner/cpuminer.go
dnldd: good catch, didn't need to move it.

dnldd: Good catch, didn't need to move it.

cmd/miner/cpuminer.go
jholdstock: My mistake, I overlooked that.

PR #154, opened 2020-02-20
build: go 1.14

dajohi:


dajohi:
https://github.com/etcd-io/bbolt/pull/201

PR #155, opened 2020-02-23
Miscellaneous minor improvements for GUI

jholdstock:
Bundled into a single PR for convenience. See individual commits for further information.

### DIFF
--- a/gui/gui.go
+++ b/gui/gui.go
@@ -150,7 +150,7 @@ func (ui *GUI) route() {
 }
 
 // renderTemplate executes the provided template.
-func (ui *GUI) renderTemplate(w http.ResponseWriter, _ *http.Request, name string, data interface{}) {
+func (ui *GUI) renderTemplate(w http.ResponseWriter, name string, data interface{}) {
 	var doc bytes.Buffer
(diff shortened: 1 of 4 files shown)


gui/gui.go
```diff
@@ -165,6 +165,18 @@ func (ui *GUI) renderTemplate(w http.ResponseWriter, _ *http.Request, name strin
 	}
 }
 
+func getSession(r *http.Request, cookieStore *sessions.CookieStore) (*sessions.Session, error) {
+	session, err := cookieStore.Get(r, "session")
+	if err != nil {
+		// "value is not valid" occurs if the CSRF secret changes.
+		// This is common and not a cause for concern - no need to error out.
```
dnldd: This doesn't look right, in production the CSRF secret isn't expected to change once set and persisted. The harness generates a new one every time it's started however.

jholdstock: That matches my understanding, and that is why I removed it - in production this should never occur, and in dev/testing it is just noise.

I can add it back, but how do you feel about dropping it from `Error` to `Warn`?

dnldd: Yeah `Warn` works, let's do that.

jholdstock: Done

PR #157, opened 2020-02-24
Admin login through AJAX

jholdstock:
Marked as draft until #155 is merged (will likely need a rebase).

This PR uses AJAX to submit the admin login form, smoothing the UX when an incorrect password is supplied or an error occurs (eg. server is not contactable, no network).

Requires an upgrade from jquery-slim to full jquery.

Also focus on text inputs automatically when modals are opened, reducing the need for extra mouse clicks.


jholdstock:
Copying feedback from Matrix:

> I particularly don't like where the account info error ("unable to generate account id") is showing up since it pushes the ui on the page lower. Also the error in the modal would be better presented below the input I think. Let's try keeping to modal around and showing the error in it, below the input.

> I think the errors should display in the modals that generate them

> Probably need to move the buttons a bit to make space for it beneath the input if they're are sharing the same line

> Try putting the error beneath the buttons as well so no UI element moves around when an error shows up

> It's either that or we find a place on the page for errors, that'd mean modals will close and the error will be shown

gui/index.go
```diff
@@ -142,3 +142,35 @@ func (ui *GUI) Homepage(w http.ResponseWriter, r *http.Request) {
 
 	ui.renderTemplate(w, "account", data)
 }
+
+// DoesAccountExist is the handler for "GET /account". If the provided address
+// has an account on the server a "200 OK" response is returned, otherwise a
+// "400 Bad Request" is returned.
+func (ui *GUI) DoesAccountExist(w http.ResponseWriter, r *http.Request) {
```
dnldd: Rename this to `IsPoolAccount` instead, it's a better name.

gui/index.go
```diff
@@ -127,12 +127,6 @@ func (ui *GUI) Homepage(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if len(work) == 0 && len(payments) == 0 {
```
dnldd: Why is this check no longer needed?

dnldd:
Looks good overall, nice work. Should be good to go once the issues highlighted are resolved.

gui/index.go
jholdstock: Done

gui/index.go
jholdstock: I don't think this check was ever required. Removing the check has no negative consequences, but it allows users to check their connected clients or lookup their account ID regardless of whether they have any mined work or payments.

dnldd: Ah yeah this would prevent the page from loading, the use cases you've stated are valid.

PR #160, opened 2020-03-19
pool: refactor & improve chainstate test coverage.

dnldd:
This fixes a bug with the chainstate where disconnected mined blocks were deleted instead of being unconfirmed. Disconnected blocks are also used to unconfirm mined work if the mined work is the associated parent. Additional tests have been added accordingly.

### DIFF
--- a/pool/chainstate.go
+++ b/pool/chainstate.go
@@ -28,6 +28,14 @@ type ChainStateConfig struct {
 	GeneratePayments func(uint32, dcrutil.Amount) error
 	// GetBlock fetches the block associated with the provided block hash.
 	GetBlock func(*chainhash.Hash) (*wire.MsgBlock, error)
+	// PruneJobs removes all jobs with heights less than the provided height.
+	PruneJobs func(*bolt.DB, uint32) error
+	// PruneAcceptedWork removes all accepted work not confirmed as mined
+	// work with heights less than the provided height.
+	PruneAcceptedWork func(*bolt.DB, uint32) error
+	// PendingPaymentsAtHeight fetches all pending payments at
+	// the provided height.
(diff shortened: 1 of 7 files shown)


pool/chainstate.go
```diff
@@ -102,7 +110,6 @@ func (cs *ChainState) handleChainUpdates(ctx context.Context) {
 			if err != nil {
 				log.Errorf("unable to create header from bytes: %v", err)
 				close(msg.Done)
-				cs.cfg.Cancel()
```
jholdstock: While it is quite well explained in the test file, it's really not clear from just reading the source code here why some of the error cases are calling `Cancel()` and some of them are not. (For example - `PruneJobs()` error causes `Cancel()` to be called, and `AcceptedWorkID()` error does not.)

It might be worth explaining each case here as well, so a casual reader can understand without having to read both files.

pool/chainstate.go
```diff
@@ -190,32 +193,52 @@ func (cs *ChainState) handleChainUpdates(ctx context.Context) {
...
 			}
 
-			// Delete mined work if it is disconnected from the chain. At this
-			// point a mined confirmed block will have its corresponding
-			// accepted block record persisted.
-			id := AcceptedWorkID(header.BlockHash().String(), header.Height)
+			// Check if the disconnected block confirms a mined block, if it
+			// does unconfirm it.
+			id := AcceptedWorkID(header.PrevBlock.String(), header.Height-1)
+			confWork, _ := FetchAcceptedWork(cs.cfg.DB, id)
```
jholdstock: I would suggest a comment here explaining why it is acceptable to ignore errors from `FetchAcceptedWork()`

pool/chainstate.go
dnldd: Thanks, I started combing through some more and found other cases that need better handing.

PR #158, opened 2020-03-12
multi: add obelisk DCR1 support.

dnldd:
This resolves #110.

Issue #110, opened 2019-07-22
Add support for Obelisk DCR1


dnldd:
The plan is to implement DCR1 support after v1 is released.

PR #161, opened 2020-03-20
Add Obelisk DCR1 to gui.

jholdstock:
This can be merged before #158 without causing any errors, but if it is, the DCR1 port will not be displayed in the GUI until 158 goes in.

It's unfortunate that the carousel height needs to be manually incremented when a the text inside it changes. Its a limitation inherited from the carousel code taken out of the VSP site. I will see if we can improve that.

(Part of #110)

### DIFF
--- a/gui/assets/public/css/dcrpool.css
+++ b/gui/assets/public/css/dcrpool.css
@@ -7616,22 +7616,22 @@ input[type="submit"]:disabled {
         min-height: 190px; } }
     @media (min-width: 992px) {
       .main-carousel__image-container {
-        min-height: 275px; } }
+        min-height: 300px; } }
     .main-carousel__image-container img {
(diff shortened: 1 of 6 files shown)


dnldd:
Noticed in the UI the DCR1 port isn't being shown, I think this is because this PR has not been rebased off of the DCR1 PR. Please do that to have access to all the needed info. Also reduce the number of decimal places for the hash rate from 4 to 2.

jholdstock:
I've made the suggested changes and also pushed another couple of minor fixes for mobile responsiveness.

dnldd:
Looks good but it seems there's something off with the rebase, `build: go 1.14 (#154)` which is already merged sneaked in there. I'll get the DCR1 PR finalized for merging, this will follow after.

PR #163, opened 2020-03-24
Pagination for mined blocks table

jholdstock:
PR marked as draft as it is mostly a proof-of-concept and still needs some work.

Rebased on top of #157.

Implements pagination for the mined blocks table on the homepage. Does not include any rate limiting or caching. The client code does not include any error handling, or handling for the case where no blocks have yet been mined by the pool. There is also no loading message or spinner.

The code in `accepted_work.go` could use some optimisation, because it is currently performing two database transactions where one could potentially be enough.


dnldd:
Yeah you're right, the code in `accepted_work.go` can be improved. We should use one db transaction to fetch all confirmed work and based on the offset requested the data returned can be sliced and returned. The returned set of confirmed work can be then be cached, this will be invalidated when a new confirmed work is persisted to the db. It'd be a lot more efficient if the listing call didn't even hit the db at all, it'd just use the most recent cache of confirmed work available and slice it accordingly per the offset and page size.

jholdstock:
Complete and ready for review.

- Code is now only doing one database transaction.
- Caching and rate limiting are now implemented.
- Code to handle "no mined blocks" is in place.
- Works well without Javascript. If Javascript is disabled, only the latest 10 blocks are shown.

gui/gui.go
```diff
@@ -341,56 +402,10 @@ func (ui *GUI) Run(ctx context.Context) {
...
-						})
-					}
-
-					// Update mined work cache.
-					ui.minedWorkMtx.Lock()
-					ui.minedWork = workData
-					ui.minedWorkMtx.Unlock()
-
-					quotas, err := ui.cfg.FetchWorkQuotas()
+					err := updateCachedData()
```
dnldd: I think updating cached data should be triggered by incoming new data, would save a lot of db queries.  for example if a new mined work is getting persisted it should trigger a cached update. Perhaps the caches should reside with the pool, they'll be exposed to the gui via the config struct as usual.

dnldd:
I realize my comments are improvements on the current system in place, will merge  and improve the caching system in other PRs. Nicely done.

PR #164, opened 2020-03-24
pool: improve test coverage for bolt db.

dnldd:
This adds tests for all bolt db bucket helpers as well as coverage for negative db initialization paths.
The pool database is now backed up before shutdown.

### DIFF
--- a/pool/db.go
+++ b/pool/db.go
@@ -52,6 +52,8 @@ var (
 	csrfSecret = []byte("csrfsecret")
 	// poolFeesK is the key used to track pool fee payouts.
 	poolFeesK = "fees"
+	// backup is the database backup file name.
+	backupFile = "backup.kv"
 )
 
 // openDB creates a connection to the provided bolt storage, the returned
@@ -225,13 +227,11 @@ func InitDB(dbFile string, isSoloPool bool) (*bolt.DB, error) {
 
 	if switchMode {
 		// Backup the current database and wipe it.
-		now := time.Now().Format(time.RFC3339)
-		file := fmt.Sprintf("dcrpool@%v.kv", now)
-		err := backup(db, file)
+		err := backup(db, backupFile)
 		if err != nil {
 			return nil, err
 		}
-		log.Infof("Pool mode changed, database backup %s created.", file)
+		log.Infof("Pool mode changed, database backup created.")
 		err = purge(db)
 		if err != nil {
 			return nil, err
--- a/pool/db.go
+++ b/pool/db.go
@@ -7,6 +7,7 @@ package pool
 import (
 	"encoding/binary"
 	"fmt"
+	"path/filepath"
 	"time"
 
 	bolt "go.etcd.io/bbolt"
@@ -227,7 +228,8 @@ func InitDB(dbFile string, isSoloPool bool) (*bolt.DB, error) {
 
 	if switchMode {
 		// Backup the current database and wipe it.
-		err := backup(db, backupFile)
+		backupPath := filepath.Join(filepath.Dir(db.Path()), backupFile)
+		err := backup(db, backupPath)
 		if err != nil {
 			return nil, err
 		}
--- a/pool/db_test.go
+++ b/pool/db_test.go
@@ -4,6 +4,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/decred/dcrd/chaincfg/v2"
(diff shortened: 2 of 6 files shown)


pool/pool_test.go
```diff
@@ -71,7 +72,11 @@ func teardownDB(db *bolt.DB, dbPath string) error {
 	if err != nil {
 		return err
 	}
-	return nil
+	backup := filepath.Join(filepath.Dir(db.Path()), backupFile)
+	if _, err := os.Stat(backup); os.IsNotExist(err) {
+		return nil
+	}
+	return os.Remove(backup)
```
jholdstock: Maybe I am misunderstanding, but this seems dangerous. Why should _test_ code by removing a backup file which is created by _production_ code? 

This could potentially delete an important backup of a production database, which is a very drastic consequence of just running some tests.

pool/db.go
```diff
@@ -225,13 +227,11 @@ func InitDB(dbFile string, isSoloPool bool) (*bolt.DB, error) {
...
 		// Backup the current database and wipe it.
-		now := time.Now().Format(time.RFC3339)
-		file := fmt.Sprintf("dcrpool@%v.kv", now)
-		err := backup(db, file)
+		err := backup(db, backupFile)
 		if err != nil {
 			return nil, err
 		}
-		log.Infof("Pool mode changed, database backup %s created.", file)
+		log.Infof("Pool mode changed, database backup created.")
```
jholdstock: I really think we should log the path and filename where the backup was created. Perhaps we could add it to `backup()` in `db.go` to ensure it is always logged and to prevent duplication?

pool/db.go
```diff
@@ -52,6 +52,8 @@ var (
 	csrfSecret = []byte("csrfsecret")
 	// poolFeesK is the key used to track pool fee payouts.
 	poolFeesK = "fees"
+	// backup is the database backup file name.
+	backupFile = "backup.kv"
```
jholdstock: Using a single hard-coded backup name creates a situation where an important backup could be overwritten and lost. Why are we no longer generating unique filenames dynamically?

This is especially risky if we are overwriting the backup file every time dcrpool is stopped

pool/pool_test.go
dnldd: I think you are misunderstanding. The test code runs from the project directory, the backup created is within the project directory as a result, the backup deletion is clearing files remaining after running tests. This has nothing to do with production. If you notice the back up is created using the path of the db, in production the backup will be created where the db resides per the configuration file, not in the project directory.

pool/db.go
dnldd: Not logging the path of the backup because the backup file location will be consistent, it'll always be created where the db resides. Probably should state this in th README as well.

pool/db.go
dnldd: There is still the option to backup the db via the admin page, this backup file is what the pool generates automatically. Also it's unlikely the backup gets corrupted during an overwrite because bolt db updates are transactional and are rolled back if there is an error.

pool/pool_test.go
jholdstock: Understood. My confusion come from them both using the same `backupFile` var, but that is only the file name, and now I notice that the path is different.

pool/db.go
jholdstock: Understood.

pool/db_test.go
```diff
@@ -24,7 +221,7 @@ func testInitDB(t *testing.T) {
 	}
 	defer td()
 
-	// Ensure the db bucket have been created.
+	// Ensure the db buckets have been created.
 	err = db.View(func(tx *bolt.Tx) error {
```
jholdstock: This error is never checked.

pool/db_test.go
```diff
@@ -57,6 +254,30 @@ func testInitDB(t *testing.T) {
 		}
 		return nil
 	})
+
+	// Persist the pool mode.
+	err = db.Update(func(tx *bolt.Tx) error {
```
jholdstock: This error is never checked

pool/db_test.go
```diff
@@ -57,6 +254,30 @@ func testInitDB(t *testing.T) {
...
+		return pbkt.Put(soloPool, b)
+	})
+
+	err = db.Close()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Trigger a pool mode switch.
+	db, err = InitDB(dbPath, true)
```
jholdstock: This error is checked twice

PR #166, opened 2020-03-26
pool: improve test coverage for payments.

dnldd:
This adds additional tests to payments and endpoints, and fixes a bug with the rate limiter where an unknown client type is passed to `withinLimit`.

### DIFF
--- a/pool/limiter.go
+++ b/pool/limiter.go
@@ -5,6 +5,7 @@
 package pool
 
 import (
+	"fmt"
 	"sync"
 
 	"golang.org/x/time/rate"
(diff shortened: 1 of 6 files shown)


pool/limiter.go
```diff
@@ -91,5 +91,10 @@ func (r *RateLimiter) withinLimit(ip string, clientType int) bool {
 		// create a new limiter if the incoming request is from a new client.
 		reqLimiter = r.addRequestLimiter(ip, clientType)
 	}
+
+	if reqLimiter == nil {
```
jholdstock: I'm unsure how I feel about this. I think the code would be cleaner if `addRequestLimiter()` didn't return nil when it is passed an unknown client type, but instead returned an error.

To my eye, the way this code is currently written looks more like Java than Go.

PR #171, opened 2020-03-29
multi: refactor connection creation out of hub.

dnldd:
This refactors establishing node and wallet connections out of the hub which ultimately makes it easier to test. The hub's configuration struct has been updated to remove now unnecessary details.

### DIFF
--- a/dcrpool.go
+++ b/dcrpool.go
@@ -143,7 +143,7 @@ func newPool(cfg *config) (*miningPool, error) {
 	p.hub.SetNodeConnection(nodeConn)
 
 	// Establish a connection to the wallet if the pool is mining as a
-	// publicly available mining poolin solo.
+	// publicly available mining pool.
 	if !cfg.SoloPool {
 		creds, err := credentials.
 			NewClientTLSFromFile(cfg.WalletRPCCert, "localhost")
--- a/dcrpool.go
+++ b/dcrpool.go
@@ -19,6 +19,9 @@ import (
 	"github.com/decred/dcrd/rpcclient/v5"
 	"github.com/decred/dcrpool/gui"
 	"github.com/decred/dcrpool/pool"
+	"github.com/decred/dcrwallet/rpc/walletrpc"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
 )
 
 // miningPool represents a decred Proof-of-Work mining pool.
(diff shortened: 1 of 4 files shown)


dcrpool.go
```diff
@@ -123,7 +123,51 @@ func newPool(cfg *config) (*miningPool, error) {
...
+	// Establish a connection to the wallet if the pool is mining as a
+	// publicly available mining poolin solo.
+	if !cfg.SoloPool {
+		creds, err := credentials.
+			NewClientTLSFromFile(cfg.WalletRPCCert, "localhost")
+		if err != nil {
+			return nil, err
+		}
+
+		grpc, err := grpc.Dial(cfg.WalletGRPCHost,
```
jholdstock: As far as I can tell, this grpc client can be [safely accessed concurrently](https://github.com/grpc/grpc-go/blob/master/Documentation/concurrency.md) and does not need to be protected by a mutex.

I believe the same is also true for the dcrd client, but I'm not 100% sure on that.

pool/hub.go
```diff
@@ -63,16 +63,32 @@ var (
...
 
+// WalletConnection defines the functionality needed by a wallet
+// grpc connection for the pool.
+type WalletConnection interface {
+	ConstructTransaction(context.Context, *walletrpc.ConstructTransactionRequest, ...grpc.CallOption) (*walletrpc.ConstructTransactionResponse, error)
+	SignTransaction(context.Context, *walletrpc.SignTransactionRequest, ...grpc.CallOption) (*walletrpc.SignTransactionResponse, error)
+	PublishTransaction(context.Context, *walletrpc.PublishTransactionRequest, ...grpc.CallOption) (*walletrpc.PublishTransactionResponse, error)
+}
+
+// NodeConnection defines the funcationality needed a mining node
```
jholdstock:
```suggestion
// NodeConnection defines the functionality needed by a mining node
```

pool/hub.go
```diff
@@ -109,6 +126,21 @@ type Hub struct {
 	wg             *sync.WaitGroup
 }
 
+// SetNodeConnection sets the mining node connection.
```
jholdstock: Could the client connections be passed as parameters to `pool.NewHub()`?

That would allow us to remove the setter functions, and also we could check `if nodeConn==nil` just once in `newHub()` - we wouldn't have to duplicate the check in `getBlock()`, `getWork()` and `submitWork()`.

dnldd: The notification handlers require an initialized hub to create in `CreateNotificationHandlers()`.

PR #173, opened 2020-03-30
gui: Implement new style errors.

jholdstock:
This PR implements the new error modals as defined by @MariaPleshkova in https://xd.adobe.com/view/a814493c-193a-4dfc-5735-0e6d459a6924-de25/

Also...
- Account details page is now on its own path `/account/{address}`. Previously it made sense to have the account details and index page on the same path because they were two very similar pages, but now they have diverged quite significantly (eg. they each have their own templates).
- The old path to check if an account exists has been renamed from `/account` to `/account_exist`
- The modal backdrop is now the correct color as per the design spec.

### DIFF
--- a/gui/account.go
+++ b/gui/account.go
@@ -1,0 +1,125 @@
+// Copyright (c) 2020 The Decred developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package gui
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/decred/dcrpool/pool"
+	"github.com/gorilla/csrf"
+	"github.com/gorilla/mux"
+)
+
+// accountPageData contains all of the necessary information to render the
+// account template.
+type accountPageData struct {
+	HeaderData       headerData
+	MinedWork        []*pool.AcceptedWork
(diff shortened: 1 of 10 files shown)


dnldd:
Testing this right now, not seeing [this modal](https://xd.adobe.com/view/a814493c-193a-4dfc-5735-0e6d459a6924-de25/) show up when I search for fake address. I actually prefer the simple `invalid address` with the red underline that is currently there. Perhaps we should keep that?

dnldd:
This is what I'm seeing when i disable js:
no-js

I've synced the PR and rebuilt twice locally, let me know if I'm doing anything wrong here.

jholdstock:
The new modal is only used if you directly visit a URL which has an issue - for example `/account/NotARealAddress`. In this case, it is not practical to show the error under the text input.

When you are searching for an account using the AJAX form, the errors should be displayed as before. This PR does not change that.

This PR also does not change the Javascript warning - that will need some extra thought because it cannot be implemented as a modal - drawing a modal depends on Javascript. However, I am not able to recreate your left-aligned Javascript warning. On my screen that warning is centred and looks better.

Issue #167, opened 2020-03-29
GUI caching optimisation


jholdstock:
Data cached for the GUI includes:

- Total pool hash rate
- Confirmed blocks mined by pool
- Work quotas describing reward distribution for accounts

The whole cache is currently refreshed once every 15 seconds. This is acceptable for the hash rate as it changes often, and it is simply a variable held in memory, but the other data is pulled from the database and should not be polled so often. Instead, we should only update the cache whenever the underlying data is changed.

dnldd:
This has been resolved fully by #183, #185, #187 and #195.

PR #174, opened 2020-03-30
Pagination for blocks mined by account

jholdstock:
Rebased on #173 

This PR implements pagination for mined blocks on the account details page. Also moves the GUI caching code into its own file which means we don't need to be concerned with the mutex handling outside of this file. Should also make #167 easier.


jholdstock:
Closing - these changes will be included in #180 instead

PR #176, opened 2020-03-31
pool: Only pay dividends if not in solo pool mode

jholdstock:


dnldd:
Good catch, in solo pool mode `PayDividends` returns without an error since there are no eligible payments to process. This is more efficient however.

PR #177, opened 2020-03-31
Config improvements

jholdstock:
- Remove default dcrd user/pass and provide a descriptive error if they are not set.
- Remove default `WalletPass` and `PoolFeeAddr`. They were set to `""`, which is already the zero value for a string.
- Remove an errant `if` statement which was preventing the default config file from loading
- Ensure `PoolFee` is `>=0` and `<=1` in solo mode. **Question: Do we actually want to allow 0% or 100%?**
- Ensure `WalletPass` is provided when in solo mode.
- Ensure `PoolFeeAddrs` is provided when in solo mode, and also ensure it is properly split when multiple addresses are provided.
- If the pool cannot start, log error to stderr and finished with exit code 1.
- Automatically use the default main/test/sim network ports for dcrd/dcrwallet

### DIFF
--- a/config.go
+++ b/config.go
@@ -36,8 +36,8 @@ const (
 	defaultDBFilename            = "dcrpool.kv"
 	defaultTLSCertFilename       = "dcrpool.cert"
 	defaultTLSKeyFilename        = "dcrpool.key"
-	defaultDcrdRPCHost           = "127.0.0.1:19109"
-	defaultWalletGRPCHost        = "127.0.0.1:51028"
+	defaultDcrdRPCHost           = "127.0.0.1"
+	defaultWalletGRPCHost        = "127.0.0.1"
 	defaultMaxGenTime            = time.Second * 15
 	defaultPoolFee               = 0.01
 	defaultLastNPeriod           = time.Hour * 24
@@ -115,7 +115,7 @@ type config struct {
 	DCR1Port              uint32        `long:"dcr1port" ini-name:"dcr1port" description:"Obelisk DCR1 connection port."`
 	poolFeeAddrs          []dcrutil.Address
 	dcrdRPCCerts          []byte
-	net                   *chaincfg.Params
+	net                   *params
 }
 
 // serviceOptions defines the configuration options for the daemon as a service on
@@ -289,6 +289,16 @@ func cleanAndExpandPath(path string) string {
 	}
 
 	return filepath.Join(homeDir, path)
+}
+
+// normalizeAddress returns addr with the passed default port appended if
+// there is not already a port specified.
+func normalizeAddress(addr, defaultPort string) string {
+	_, _, err := net.SplitHostPort(addr)
+	if err != nil {
+		return net.JoinHostPort(addr, defaultPort)
+	}
+	return addr
 }
 
 // loadConfig initializes and parses the config using a config file and command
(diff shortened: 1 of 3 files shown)


config.go
```diff
@@ -528,16 +542,20 @@ func loadConfig() (*config, []string, error) {
...
+		cfg.net = &testNet3Params
 	case chaincfg.SimNetParams().Name:
-		cfg.net = chaincfg.SimNetParams()
+		cfg.net = &simNetParams
 	default:
 		return nil, nil, fmt.Errorf("unknown network provided %v",
 			cfg.ActiveNet)
 	}
 
+	// Add default ports for the active network if there's no ports specified
```
dnldd: s/ there's / there are

config.go
```diff
@@ -546,6 +564,30 @@ func loadConfig() (*config, []string, error) {
 			return nil, nil, err
 		}
 
+		// Ensure pool fee is valid
+		if cfg.PoolFee < 0 || cfg.PoolFee > 1 {
```
dnldd: This is fine. I think it makes sense to allow pools to set the bounds of the pool fee (0% or 100%), it is up to the user to determine if the pool fee being asked for is reasonable and make a decision on that.

config.go
```diff
@@ -546,6 +564,30 @@ func loadConfig() (*config, []string, error) {
...
+			str := "%s: poolfee should be between 0 and 1"
+			err := fmt.Errorf(str, funcName)
+			return nil, nil, err
+		}
+
+		// Ensure the password to unlock the wallet.
+		// Needed to pay dividends to pool contributors.
```
dnldd: The comments here seem incomplete.

config.go
```diff
@@ -546,6 +564,30 @@ func loadConfig() (*config, []string, error) {
...
+		if cfg.WalletPass == "" {
+			str := "%s: the walletpass option is not set"
+			err := fmt.Errorf(str, funcName)
+			return nil, nil, err
+		}
+
+		// Ensure address to collect pool fees is provided.
+		if len(cfg.PoolFeeAddrs[0]) == 0 {
```
dnldd: We can refine this some more, this should only be set when not in solo pool mode.

config.go
```diff
@@ -546,6 +564,30 @@ func loadConfig() (*config, []string, error) {
...
+			str := "%s: the poolfeeaddrs option is not set"
+			err := fmt.Errorf(str, funcName)
+			return nil, nil, err
+		}
+
+		// Parse pool fee addresses.
+		cfg.PoolFeeAddrs = strings.Split(cfg.PoolFeeAddrs[0], ",")
```
dnldd: This would only pick the first address in the list of addresses for pool fees. Parsing pool fee addresses is already handled here: https://github.com/decred/dcrpool/blob/master/config.go#L549-L560

jholdstock: Actually I spent quite some time on this and I am sure it is right. It's something we discovered in dcrstakepool. `jessevdk/go-flags` does not automatically split the string into an array. It instead simply puts the provided value in the first item of the array, regardless of the value. For example:

```
input: ""
parsed to: [""]

input: "1"
parsed to: ["1"]

input: "1,2,3"
parsed to: ["1,2,3"]
```

The array always contains one item (hence the usage of `[0]`) and it needs to be split manually using `strings.Split()`

config.go
dnldd: Also this is taking the length of the first pool fee address, not the length of addresses provided.

config.go
jholdstock: Proof of concept: https://github.com/jholdstock/flags-poc

dnldd: Can confirm this behaviour by `jessevdk/go-flags` is correct.

config.go
dnldd: in light of the issue discovered with `jessevdk/go-flags` please add comments explaining why things are being done this way for future reference.

PR #179, opened 2020-04-07
Use wallet gRPC

jholdstock:
- Use correct gRPC ports for dcrwallet.
- `cmd/miner` - log an error on invalid network selection
- Ensure difficulty is not zero - return 1 instead.

### DIFF
--- a/pool/message.go
+++ b/pool/message.go
@@ -349,14 +349,16 @@ func ParseSubscribeResponse(resp *Response) (string, string, string, uint64, err
 // SetDifficultyNotification creates a set difficulty notification message.
 func SetDifficultyNotification(difficulty *big.Rat) *Request {
 	diff, _ := difficulty.Float64()
-	iDiff := uint64(diff)
-	if iDiff == 0 {
-		iDiff = 1
+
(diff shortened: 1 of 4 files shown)


pool/message.go
```diff
@@ -349,9 +349,14 @@ func ParseSubscribeResponse(resp *Response) (string, string, string, uint64, err
 // SetDifficultyNotification creates a set difficulty notification message.
 func SetDifficultyNotification(difficulty *big.Rat) *Request {
 	diff, _ := difficulty.Float64()
+	iDiff := uint64(diff)
```
dnldd, line 352: Could you use a more descriptive name instead of `iDiff` here?

jholdstock: Done, also added a little comment to explain.

dnldd:
The comment `Convert the diff to a uint64 to ensure it does not get rounded to zero`. isn't accurate because it's the casting of the diff to uint64 that's doing the rounding/truncation. The check there is making sure the diff value is never zero. Please update the comment.

jholdstock:
Done

PR #180, opened 2020-04-08
Introduce cache and more pagination.

jholdstock:
As discussed on Matrix, bundling changes from #173 and #174 into a new PR with some additional changes.

### New style errors

This PR implements the new error modals as defined by @MariaPleshkova in https://xd.adobe.com/view/a814493c-193a-4dfc-5735-0e6d459a6924-de25/. These errors are displayed when an invalid URL is visited, for example `/account?address=NotARealAddress`. When you are searching for an account using the AJAX form, the errors should be displayed as before, this PR does not change those.

**Note:** This PR does not change the Javascript warning - that will need some extra thought because it cannot be implemented as a modal - drawing a modal depends on Javascript.

### Cache

`Cache` has been introduced, which acts as a data buffer between hub and gui. Cache has setter functions which are used by the hub, and getter functions which are used by gui. It holds pre-formatted data which is ready to be served directly to gui without needing to lock any mutexes in the hub or hit the database. This facilitates implementing #167 because now the hub is able to push data into the cache when it knows the data has changed, whereas the previous model had the gui pulling data from the hub on a timer, with no knowledge of whether the underlying data had changed or not.

### Pagination

Pagination has been added for both the Mined Blocks and the Connected Clients tables on the account page.

### Rate Limiting

Rate limiting moved to a middleware which removes a lot of duplicated code.

As a result, rate limiting has been added to some places it was previously missing:
- `/ws` the websocket route
- `/logout` for admin logout

The limit has been increased from 3 to 5 for GUI clients. This is because some of the pages are now making extra requests (pagination, establishing a websocket, AJAX form submissions) and the limit of 3 was causing issues.

### DIFF
--- a/gui/assets/public/css/pagination.css
+++ b/gui/assets/public/css/pagination.css
@@ -58,8 +58,8 @@
   background-position: center;
 }
 .paginationjs .pagination-arrow-right {
-  background-image: url(/images/right_arrow.svg);
+  background-image: url('/assets/images/right_arrow.svg');
 }
 .paginationjs .pagination-arrow-left {
-  background-image: url(/images/left_arrow.svg);
+  background-image: url('/assets/images/left_arrow.svg');
 }
--- a/gui/cache.go
+++ b/gui/cache.go
@@ -61,7 +61,7 @@ type Cache struct {
 
 // InitCache initialises and returns a cache for use in the GUI.
 func InitCache(work []*pool.AcceptedWork, quotas []*pool.Quota,
-	clients map[string][]*pool.ClientInfo, blockExplorerURL string) *Cache {
+	clients []*pool.Client, blockExplorerURL string) *Cache {
 
 	cache := Cache{blockExplorerURL: blockExplorerURL}
 	cache.updateMinedWork(work)
@@ -127,18 +127,18 @@ func (c *Cache) getPoolHash() string {
 
 // updateClients will refresh the cached list of connected clients, as well as
 // recalculating the total hashrate for all connected clients.
-func (c *Cache) updateClients(clients map[string][]*pool.ClientInfo) {
-	clientInfo := make(map[string][]client)
+func (c *Cache) updateClients(clients []*pool.Client) {
+	clientInfo := make(map[string][]client, 0)
 	poolHashRate := new(big.Rat).SetInt64(0)
-	for account, c2 := range clients {
-		for _, c := range c2 {
-			clientInfo[account] = append(clientInfo[account], client{
-				Miner:    c.Miner,
-				IP:       c.IP,
-				HashRate: hashString(c.HashRate),
-			})
-			poolHashRate = poolHashRate.Add(poolHashRate, c.HashRate)
-		}
+	for _, c := range clients {
+		clientHashRate := c.FetchHashRate()
+		accountID := c.FetchAccountID()
+		clientInfo[accountID] = append(clientInfo[accountID], client{
+			Miner:    c.FetchMinerType(),
+			IP:       c.FetchIPAddr(),
+			HashRate: hashString(clientHashRate),
+		})
+		poolHashRate = poolHashRate.Add(poolHashRate, clientHashRate)
 	}
 
 	c.poolHashMtx.Lock()
--- a/gui/cache.go
+++ b/gui/cache.go
@@ -11,6 +11,14 @@ import (
 	"github.com/decred/dcrpool/pool"
 )
 
+// client represents a mining client. It is json annotated so it can easily be
+// encoded and sent over a websocket or pagination request.
+type client struct {
+	Miner    string `json:"miner"`
+	IP       string `json:"ip"`
+	HashRate string `json:"hashrate"`
+}
+
 // minedWork represents a block mined by the pool. It is json annotated so it
 // can easily be encoded and sent over a websocket or pagination request.
 type minedWork struct {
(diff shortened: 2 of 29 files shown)


gui/assets/public/css/pagination.css
```diff
@@ -58,8 +58,8 @@
   background-position: center;
 }
 .paginationjs .pagination-arrow-right {
-  background-image: url(/images/right_arrow.svg);
+  background-image: url('/assets/images/right_arrow.svg');
```
dnldd, line 61: Why has the directory structure changed for gui assets?

gui/gui.go
```diff
@@ -122,35 +108,49 @@ type poolStats struct {
...
+	guiRouter := ui.router.PathPrefix("/").Subrouter()
+
+	// sessionMiddleware must be run before rateLimitMiddleware.
+	guiRouter.Use(ui.sessionMiddleware)
+	guiRouter.Use(ui.rateLimitMiddleware)
+	guiRouter.Use(csrf.Protect(ui.cfg.CSRFSecret, csrf.Secure(true)))
+
+	guiRouter.HandleFunc("/", ui.Homepage).Methods("GET")
+	guiRouter.HandleFunc("/account", ui.Account).Methods("GET")
+	guiRouter.HandleFunc("/account_exist", ui.IsPoolAccount).Methods("GET")
```
dnldd, line 141: Perhaps this route could be `/account/exist`, `/poolaccount`, `/isaccount` or `/exist`? These options conform to the route definitions structure already in place, `account_exist`, with that underscore does not.

gui/gui.go
```diff
@@ -331,62 +314,23 @@ func (ui *GUI) Run(ctx context.Context) {
...
-		ui.workQuotasMtx.Unlock()
-
-		poolHash, _ := ui.cfg.FetchPoolHashRate()
-
-		// Update pool hash cache.
-		ui.poolHashMtx.Lock()
-		ui.poolHash = hashString(poolHash)
-		ui.poolHashMtx.Unlock()
-		return nil
+	// Initalise the cache
```
dnldd, line 317: Missing comment punctuation.

pool/hub.go
```diff
@@ -562,72 +562,27 @@ func (h *Hub) Run(ctx context.Context) {
...
-type ClientInfo struct {
-	Miner    string
-	IP       string
-	HashRate *big.Rat
-}
-
 // FetchClientInfo returns connection details about all pool clients.
-func (h *Hub) FetchClientInfo() map[string][]*ClientInfo {
-	clientInfo := make(map[string][]*ClientInfo)
+func (h *Hub) FetchClientInfo() []*Client {
```
dnldd, line 566: If you're going to maintain the modifications to this function then the function signature will have to change some more. Since it's returning a slice of clients now I don't think it should be called `FetchClientInfo` anymore, perhaps `FetchHubClients` or `FetchClients` would do. Also the name `allClients` seems unnecessary here, you could call the client slice `clients` and enumerate the endpoint with `for _, c := range endpoint.clients`, this way you can append to the slice like this: `clients = append(clients, c)`. I think this is readable without being verbose.

pool/acceptedwork_test.go
```diff
@@ -93,27 +93,14 @@ func testAcceptedWork(t *testing.T, db *bolt.DB) {
 			fetchedWork.Height, workC.Height)
 	}
 
-	// Ensure requesting negative or zero most recent accepted work returns
-	// an error.
-	_, err = listMinedWorkByAccount(db, string(id), -1)
```
dnldd, line 98: Can we keep `listMinedWorkByAccount` around just for these tests?

gui/gui.go
```diff
@@ -71,26 +69,19 @@ type Config struct {
...
 	// based on work contributed per the payment scheme used by the pool.
 	FetchWorkQuotas func() ([]*pool.Quota, error)
-	// FetchPoolHashRate returns the hash rate of the pool.
-	FetchPoolHashRate func() (*big.Rat, map[string][]*pool.ClientInfo)
 	// BackupDB streams a backup of the database over an http response.
 	BackupDB func(w http.ResponseWriter) error
-	// FetchClientInfo returns connection details about all pool clients.
-	FetchClientInfo func() map[string][]*pool.ClientInfo
+	// FetchClientInfo returns details about all connected pool clients.
+	FetchClientInfo func() []*pool.Client
```
dnldd, line 80: This would have to be updated as well as the description if you make the changes suggested to the function signature.

gui/cache.go
```diff
@@ -0,0 +1,158 @@
...
+	minedWorkMtx sync.RWMutex
+
+	workQuotas    []workQuota
+	workQuotasMtx sync.RWMutex
+
+	poolHash    string
+	poolHashMtx sync.RWMutex
+
+	clients    map[string][]client
+	clientsMtx sync.RWMutex
```
dnldd, line 59: Don't think it's necessary to group struct fields like this, I think th code base only has atomically updated fields separated from the rest.

dnldd:
Done reviewing.

gui/assets/public/css/pagination.css
jholdstock: This is done to simplify the `route()` function in `gui.go`:

Screenshot from 2020-04-16 10-04-47

Using a common root directory for all assets means this code only has to appear once, otherwise all of this code needs to be duplicated for each directory (3x for `images`, `js`, `css`). 

If we were to add further directories in future, that code would have to be duplicated again.

Further, I expect we will want to add extra code for this router in future (eg. `cache-control` HTTP headers), and that would also need to be duplicated 3 times.

gui/gui.go
jholdstock: Good catch. Reading further on this topic I believe a more proper solution would be to use the same `/account` URL as the page itself, but change the request type to `HEAD`.

https://stackoverflow.com/questions/21087913/proper-route-for-checking-resource-existence-in-a-restful-api
also links through to https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html (section 9.4 is useful)

pool/acceptedwork_test.go
jholdstock: `listMinedWorkByAccount` was called three times in test code.

Two of these appearances were just testing the behaviour of `listMinedWorkByAccount` itself, so I wont bother to reinstate those.

The third call was to ensure mined work is persisted in the database correctly, so this one seems worth reinstating. I will put back some code which performs this check.

gui/cache.go
jholdstock: You mean the new lines? I included those just for the sake of readability and making it easier to see that each field has its own mutex, but I am happy to remove them.

jholdstock:
Thanks @dnldd, I've pushed commits implementing review changes.

gui/assets/public/css/pagination.css
dnldd: noted, thanks.

pool/hub.go
```diff
@@ -562,17 +562,17 @@ func (h *Hub) Run(ctx context.Context) {
 	h.shutdown()
 }
 
-// FetchClientInfo returns connection details about all pool clients.
-func (h *Hub) FetchClientInfo() []*Client {
-	allClients := make([]*Client, 0)
+// FetchClients returns connection details about all pool clients.
```
dnldd, line 565: // FetchClients returns all connected pool clients.

gui/gui.go
```diff
@@ -76,8 +76,8 @@ type Config struct {
 	FetchWorkQuotas func() ([]*pool.Quota, error)
 	// BackupDB streams a backup of the database over an http response.
 	BackupDB func(w http.ResponseWriter) error
-	// FetchClientInfo returns details about all connected pool clients.
-	FetchClientInfo func() []*pool.Client
+	// FetchClients returns details about all connected pool clients.
```
dnldd, line 79: // FetchClients returns all connected pool clients.

dnldd:
Minor comment updates and we're good to go @jholdstock

jholdstock:
Done

PR #183, opened 2020-04-17
multi: add cache refresh signalling.

dnldd:
This sets up cache refresh signals from the pool to the gui.

### DIFF
--- a/gui/gui.go
+++ b/gui/gui.go
@@ -347,27 +347,18 @@ func (ui *GUI) Run(ctx context.Context) {
 	// Use a ticker to periodically update cached data and push updates through
 	// any established websockets
 	go func(ctx context.Context) {
-		var ticks uint32
 		signalCh := ui.cfg.FetchCacheChannel()
-		ticker := time.NewTicker(5 * time.Second)
+		ticker := time.NewTicker(15 * time.Second)
 		defer ticker.Stop()
 
 		for {
 			select {
 			case <-ticker.C:
-				ticks++
-
-				// Periodically update the connected clients and pool hash rate
-				// after three ticks (15 seconds).
-				if ticks == 3 {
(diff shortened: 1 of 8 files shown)


gui/gui.go
```diff
@@ -334,8 +335,8 @@ func (ui *GUI) Run(ctx context.Context) {
 	// Use a ticker to periodically update cached data and push updates through
 	// any established websockets
```
jholdstock, line 336: Comment should be updated because this goroutine is now doing more than just running the ticker and updating websockets.

pool/hub.go
```diff
@@ -50,9 +50,40 @@ const (
...
+	// NewParent is the reason given when a work notification is generated
+	// because there is a new chain tip.
 	NewParent = "newparent"
-	NewVotes  = "newvotes"
-	NewTxns   = "newtxns"
+	// NewVotes is the reason given when a work notification is generated
+	// because new votes were received.
+	NewVotes = "newvotes"
+	// NewTxns is the reason given when a work notification is generated
+	// new transactions were received
```
jholdstock, line 60:
```suggestion
	// because new transactions were received.
```

pool/hub.go
```diff
@@ -50,9 +50,40 @@ const (
...
+const (
+	// Confirmed indicates an accepted work has been updated as
+	// confirmed mined.
+	Confirmed CacheUpdateEvent = iota
+
+	// Unconfirmed indicates a previously confimed mined work has been
+	// updated to unconfirmed due to a reorg.
+	Unconfirmed
+
+	// ClientUpdate indicates a new client has connected to the pool.
```
jholdstock, line 77:
```suggestion
	// ConnectedClient indicates a new client has connected to the pool.
```

Issue #181, opened 2020-04-08
Inform user when not all data is displayed with no-js


jholdstock:
Pagination does not work when Javascript is disabled, and tables are limited to displaying only the 10 most recent items. We should display some indication for the user that they are only seeing a subset of the available data.

PR #185, opened 2020-04-18
Paginate Next Reward Payment

jholdstock:
Pagination for "Next Reward Payment" on the homepage.


gui/cache.go
```diff
@@ -94,6 +95,13 @@ func (c *Cache) getMinedWork() []minedWork {
 
 // updateQuotas refreshes the cached list of pending dividend payments.
 func (c *Cache) updateQuotas(quotas []*pool.Quota) {
+
+	// Sort list so the largest percentages will be shown first.
+	sort.Slice(quotas, func(i, j int) bool {
+		larger := quotas[i].Percentage.Cmp(quotas[j].Percentage) > 0
```
dnldd, line 102: simpler to do `return quotas[i].Percentage.Cmp(quotas[j].Percentage) > 0`

gui/cache.go
```diff
@@ -48,8 +48,8 @@ type Cache struct {
 	blockExplorerURL string
 	minedWork        []minedWork
 	minedWorkMtx     sync.RWMutex
-	workQuotas       []workQuota
-	workQuotasMtx    sync.RWMutex
+	dividends        []dividend
```
dnldd, line 51: Hmm not sure about this change considering the values being reported here are not actual amounts being paid to participating accounts but percentages of the work contributed so far towards finding the next block. Maybe `work quota` is also not descriptive or simple enough, but using `dividend` here is misleading.

gui/cache.go
jholdstock: agreed

gui/cache.go
jholdstock: Renamed to RewardQuota as discussed on Matrix

gui/index.go
```diff
@@ -42,6 +41,12 @@ func (ui *GUI) renderIndex(w http.ResponseWriter, r *http.Request, modalError st
 		}
 	}
 
+	// Get the next reward payment percentages (max 10).
+	rewardQuotas := ui.cache.getRewardQuotas()
```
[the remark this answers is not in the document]
jholdstock: This is for the initial render. When the page is first loaded we only want to display the top ten (ie. the first page). 

When the pagination Javascript loads, clickable buttons are displayed which allow the user to load additional data on the next pages. If the Javascript does not load, only the first 10 results will be visible (hence #181).

If this is not limited, it is a potential DoS vector. We had a similar problem with VSP where information about 100s of tickets was passed into a template which caused the templating engine to take 10+ seconds to render.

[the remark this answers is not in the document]
dnldd: Gotcha, thanks.

Issue #138, opened 2019-10-31
Support payment processing requests.


dnldd:
Pool accounts should be able to request for payment processing if mature payments due them are below the minimum payment processing amount. This would be mostly used by pool accounts to claim mature unprocessed payments due them before leaving a mining pool.

dnldd:
functionality has been added for this now, needs to usable via the pool ui.

jholdstock:
If you could write up a spec for this I could probably get it implemented this afternoon/tomorrow

dnldd:
Payment processing requests should be accessible to the gui in #143 once it's in, it requires just the address as a parameter. Also only one payment processing request can be created for an account at a time.

PR #187, opened 2020-04-20
Paginate pending and archived payments

jholdstock:
This PR adds pagination to pending and archived payments displayed on the account page.

Pending payments are now displayed in their own table above the archived payments, as well as also being handled separately in the back-end.

One of the drivers for this change is that pending payments and archived payments will change at different times, and so the cache for each can be updated separately. The cache for archived payments may become pretty large, so we don't want to update it unnecessarily. I believe this separation will also be beneficial for implementing #138, when further GUI elements related to pending payments will be added (eg. a button to request payment, explainer text or notifications). If an account has no pending payments, the whole Pending Payments section can be hidden completely.


jholdstock:
Note to self - need to sort pending payments.

PR #192, opened 2020-04-21
dcrpool: align rpcuser configuration with rpcuser help message 

githubsands:
rpcuser help message states that the default user is drpc this corrects that

### DIFF
--- a/config.go
+++ b/config.go
@@ -36,6 +36,7 @@ const (
 	defaultDBFilename            = "dcrpool.kv"
 	defaultTLSCertFilename       = "dcrpool.cert"
 	defaultTLSKeyFilename        = "dcrpool.key"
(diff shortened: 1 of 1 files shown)


dnldd:
I don't think setting the rpc user name does much here since the rpc user and password come as a pair and for dcrpool should be sourced from dcrd. Setting the username does nothing but circumvent the error prompt. @jholdstock could you chime in on this please.

jholdstock:
Providing default credentials is a bad security practise. Besides - I don't see what issue this is trying to fix?

```
$ dcrpool --help | grep rpcpass
      --rpcpass=         Password for RPC connections.
```

Looks fine to me

dnldd:
Closing.

PR #193, opened 2020-04-21
[WIP] dcrpool: prompt user for adminpass if its not in configuration

githubsands:
Currently dcrpool closes if adminpass is not set. This PR sets a prompt that asks users for an adminpass rather then having to rerun the application again.

### DIFF
--- a/config.go
+++ b/config.go
@@ -514,6 +514,7 @@ func loadConfig(ctx context.Context) (*config, []string, error) {
 				continue PromptAdminPassLoop
 			}
 
+			writeToConfig(cfg.ConfigFile, "AdminPass", cfg.AdminPass)
 			// exit the while loop
 			break
 		}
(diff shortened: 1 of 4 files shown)


jholdstock:
dcrpool is a daemon and not an interactive process. It is designed to run as a background service on an unattended machine, and so all required config should be provided through the config file.

Changing dcrpool into an interactive process would need to be well justified, and I'm not I really see the benefit here. Especially when the amount of additional code and new third party dependencies are considered.

I recommend opening issues about this kind of thing so the pros/cons can be discussed before starting work.

dnldd:
I agree with @jholdstock  here, not seeing the benefits of setting the adminpass interactively. Say that was desirable why not set rpcuser and rpcpass interactively too? See the can of worms this opens up?. Please discuss the ideas and modifications with one of us or open an issue to get some feedback before proceeding to work on it. Closing.

githubsands:
@dnldd  @jholdstock Aim of this was to beable to run dcrpool simply as its self on my own machine quickly from a perspective of someone who doesn't quiet have the topology in mind of all the services required to get this organism moving.  This could be done and with it provide a with `--noinitialload` flag like done in dcrwallet.  But rather a `--initialload` that gives these prompts in this PR.

PR #201, opened 2020-04-22
gui: Only display confirmed blocks on the homepage

jholdstock:
Homepage was showing some unconfirmed blocks due to the pagination endpoint not filtering correctly.

This PR has been updated to also include some improvements to the cache. The cache now performs all filtering and paging of data, rather than the calling code needing to do it. This is slightly more memory efficient, and also reduces duplication and simplifies calling code.

### DIFF
--- a/gui/account.go
+++ b/gui/account.go
@@ -15,11 +15,11 @@ import (
 // account template.
 type accountPageData struct {
 	HeaderData       headerData
-	MinedWork        []minedWork
-	ArchivedPayments []archivedPayment
-	PendingPayments  []pendingPayment
+	MinedWork        *[]minedWork
+	ArchivedPayments *[]archivedPayment
+	PendingPayments  *[]pendingPayment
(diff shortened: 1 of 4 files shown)


gui/account.go
```diff
@@ -15,11 +15,11 @@ import (
 // account template.
 type accountPageData struct {
 	HeaderData       headerData
-	MinedWork        []minedWork
-	ArchivedPayments []archivedPayment
-	PendingPayments  []pendingPayment
+	MinedWork        *[]minedWork
```
dnldd, line 18: A slice is already a reference type, why are you taking the pointer there? Perhaps you're looking for `[]*minedWork` which is lightweight compared to `[]minedWork`.

jholdstock: Hm good point. The idea was for the cache to just return a pointer to the slice it has created, because we dont need to create a copy, but as you say this is actually the default behaviour anyway. I will change it back

Issue #205, opened 2020-04-27
Confirmed mined blocks after a reorg should not generate new payments.


dnldd:
Noticed in testing that reorged confirmation blocks when reconnected would trigger a new set of payments generated for for the coinbase of the mined block being confirmed. This is incorrect because the existing payment for the block are not deleted when the confirmation block is disconnected.

dnldd:
Resolved by #207.

Issue #204, opened 2020-04-27
Payments should reference coinbases being spent.


dnldd:
We noticed in testing v1.1.0 on testnet that input selection via the wallet's `constructTx` grpc sometimes included other coinbases.  This caused issues down the line in both cases because the included coinbase in the tx became change (confs at 1 now) after publishing the tx and no longer could satisfy the required confirmation constraint of `coinbaseMaturity+1`. This created a deficit in `prevOuts` at the required confirmations to process further payments. 

To resolve this, payments generated need to reference the coinbase they are spending so that transactions can be constructed with the referenced coinbases instead of relying on `constructTx` to select inputs for the transaction.

dnldd:
With payments directly referencing the outputs they spend, the payment manager will have to be reworked because each payout transaction will no longer be generating change. Fees will be deducted from the pool fees collected. This means `txfeereserve` will no longer be needed and will have to be removed, `maxtxfeereserve` as well. Their corresponding config variables will be deprecated and removed in v1.1.1.

dnldd:
Payment processing requests will also be removed because with the updated payment processing payments will be processed immediately they are spendable.

dnldd:
Resolved by #207.

PR #207, opened 2020-04-30
pool: fix payment process.

dnldd:
This ultimately fixes the bugs related to payment processing (#204 and #205) discovered in testing, the `paymentmgr` and `chaincfg` have also assimilated related share and payment helper funcs. The database has also undergone a number of upgrades to facilitate the implemented fixes. Associated tests of moved helpers have been updated accordingly.

Work in progress.


pool/upgrades.go
```diff
@@ -15,16 +17,26 @@ const (
 	// transactionId field to the payments struct for payment tracking purposes.
 	transactionIDVersion = 1
 
+	// shareIDVersion is the second version of the database. It updates
+	// the share id key and removes the created on time field.
+	shareIDVersion = 2
```
jholdstock, line 22: Both upgrades are being included in a single PR, so why do we need to increment the db version twice? Will every individual DB change always correspond to a new db version?

pool/upgrades.go
```diff
@@ -136,6 +148,181 @@ func transactionIDUpgrade(tx *bolt.Tx) error {
...
+		return MakeError(ErrBucketNotFound, desc, nil)
+	}
+
+	sbkt := pbkt.Bucket(shareBkt)
+	if sbkt == nil {
+		desc := fmt.Sprintf("bucket %s not found", string(shareBkt))
+		return MakeError(ErrBucketNotFound, desc, nil)
+	}
+
+	id := func(account string, createdOn int64) string {
```
jholdstock, line 177: Why is the ID function duplicated here rather than using the function from `share.go`?

In the payment upgrade, the ID function from `payment.go` is used.

pool/payment.go
```diff
@@ -14,34 +15,44 @@ import (
...
+	BlockHash string `json:"blockhash"`
+	Coinbase  string `json:"coinbase"`
+}
+
+// Payment represents value paid to a pool account or collected fees.
 type Payment struct {
 	Account           string         `json:"account"`
 	EstimatedMaturity uint32         `json:"estimatedmaturity"`
 	Height            uint32         `json:"height"`
+	Source            *PaymentSource `json:"source"`
```
jholdstock, line 29: Might be worth a comment here to explain the Source fields can possibly be empty if the payment was created in an old version of the database.

pool/chainstate.go
```diff
@@ -95,6 +93,87 @@ func (cs *ChainState) fetchCurrentWork() string {
...
+
+		toDelete := [][]byte{}
+		c := bkt.Cursor()
+		for k, _ := c.First(); k != nil; k, _ = c.Next() {
+			height, err := hex.DecodeString(string(k[:8]))
+			if err != nil {
+				return err
+			}
+
+			if bytes.Compare(height, heightBE) < 0 {
```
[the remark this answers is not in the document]
dnldd: good catch, `pruneAcceptedWork` can be more efficient comparing heights as bytes and avoiding the unint32 conversion for every entry being compared to.

pool/share_test.go
```diff
@@ -14,22 +16,27 @@ import (
...
 	}
 
 	err := share.Create(db)
 	if err != nil {
-		return fmt.Errorf("unable to persist share: %v", err)
+		return nil, fmt.Errorf("unable to persist share: %v", err)
 	}
 
-	return nil
+	return share, nil
```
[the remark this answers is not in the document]
dnldd: Will take another look but I think it's being used in commits still being worked on.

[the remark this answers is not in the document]
dnldd: Had another look and removed it, it was not necessary.

pool/payment.go
```diff
@@ -14,34 +15,44 @@ import (
...
 }
 
-// GeneratePaymentID generates a unique id using the provided account and the
-// created on nano time.
-func GeneratePaymentID(createdOnNano int64, height uint32, account string) []byte {
-	BECreatedOnNano := hex.EncodeToString(nanoToBigEndianBytes(createdOnNano))
-	id := fmt.Sprintf("%v%v", BECreatedOnNano, account)
-	return []byte(id)
+// paymentID generates a unique id using the provided payment details.
+func paymentID(height uint32, createdOnNano int64, account string) []byte {
```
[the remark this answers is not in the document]
dnldd: I disagree here, generating IDs without an instance is needed for tests like fetching a non-existent payment and generating payment ids that fall outside/inside the bounds of test ranges for example.

pool/upgrades.go
dnldd: For db upgrades it is ideal to keep them focus on a single change to make it easier to test against the change made and track bugs if there is an issue with an upgrade. We are not combining upgrades for this reason. The db versions here are tied to a single upgrade here, so yes every db upgrade will correspond to a new db version.

pool/upgrades.go
dnldd: Yeah, I missed a step there. The `shareID` should be taking the `createdOn` as a parameter, it'll be usable in both cases as a result.

pool/paymentmgr.go
```diff
@@ -459,34 +457,140 @@ func (pm *PaymentMgr) addPaymentRequest(addr string) error {
...
+				payments = append(payments, &payment)
 			}
 		}
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	pmts := make(map[uint32][]*Payment)
```
jholdstock, line 631: Why are the payments being stored in a map like this? I don't see the need for using a map with the height as the key. - the key doesnt seem to used anywhere other than tests.

pool/paymentmgr_test.go
```diff
@@ -112,6 +116,132 @@ func testPaymentMgr(t *testing.T, db *bolt.DB) {
...
+
+	set, ok := pmtSet[height+1]
+	if !ok {
+		t.Fatalf("expected pending payments at height %d to be "+
+			"mature at height %d", height+1, 28)
+	}
+
+	if len(set) != 2 {
+		t.Fatalf("expected 2 mature pending payments from "+
+			"height %d, got %d", height, len(set))
```
jholdstock, line 219: should be `height+1`

pool/paymentmgr_test.go
```diff
@@ -112,6 +116,132 @@ func testPaymentMgr(t *testing.T, db *bolt.DB) {
...
+	pmtSet, err = mgr.maturePendingPayments(27)
+	if err != nil {
+		t.Fatalf("maturePendingPayments error: %v", err)
+	}
+
+	if len(pmtSet) != 0 {
+		t.Fatalf("expected no payment sets, got %d", len(pmtSet))
+	}
+
+	// Empty the paymnts and archived payment buckets.
```
jholdstock, line 232:
```suggestion
	// Empty the payments and archived payment buckets.
```

pool/paymentmgr.go
```diff
@@ -595,71 +644,237 @@ func (pm *PaymentMgr) maturePendingPayments(height uint32) (map[uint32][]*Paymen
...
+	}
+
+	inSizes := make([]int, len(inputs))
+	for range inputs {
+		inSizes = append(inSizes, txsizes.RedeemP2PKHSigScriptSize)
+	}
+
+	outSizes := make([]int, len(outputs))
+	for range outputs {
+		inSizes = append(inSizes, txsizes.P2PKHOutputSize)
```
jholdstock, line 756: Should be appending to outsizes here

pool/paymentmgr.go
```diff
@@ -595,71 +644,237 @@ func (pm *PaymentMgr) maturePendingPayments(height uint32) (map[uint32][]*Paymen
...
 	}
 
-	txid, err := pm.cfg.PublishTransaction(pmts, *targetAmt)
+	estSize := txsizes.EstimateSerializeSizeFromScriptSizes(inSizes, outSizes,
+		txsizes.P2PKHOutputSize)
+	estFee := txrules.FeeForSerializeSize(txrules.DefaultRelayFeePerKb, estSize)
+
+	// Deduct the transaction fees from the pool fees being paid out.
+	fees := outputs[feeAddr.String()]
+	fees = fees - estFee
```
jholdstock, line 765: Do we need to handle the case where the pool fee is not big enough to cover the transaction fee? 

It is possible to set pool fee to any value over zero, for example 0.0000001%

pool/paymentmgr.go
```diff
@@ -595,71 +644,237 @@ func (pm *PaymentMgr) maturePendingPayments(height uint32) (map[uint32][]*Paymen
...
+	for {
+		// Ensure the wallet is synced to the current chain height before
+		// proceeding with payments.
+		walletHeight, err := pm.cfg.WalletBestHeight()
 		if err != nil {
-			return err
+			return fmt.Errorf("unable to fetch best wallet height: %v", err)
+		}
+
+		if height == walletHeight {
```
jholdstock, line 804: I think we want `height <= walletHeight` here. A new block may be found while the code above is running, which will result in an infinite loop

pool/paymentmgr.go
```diff
@@ -595,71 +644,237 @@ func (pm *PaymentMgr) maturePendingPayments(height uint32) (map[uint32][]*Paymen
...
 		}
 	}
-	err = pm.cfg.DB.Update(func(tx *bolt.Tx) error {
-		err = pm.persistTxFeeReserve(tx)
-		if err != nil {
-			return err
+
+	txBroadcaster := pm.cfg.FetchTxBroadcaster()
+	if txBroadcaster == nil {
+		return fmt.Errorf("tx broadcastor unset")
```
jholdstock, line 811:
```suggestion
		return fmt.Errorf("tx broadcaster unset")
```

pool/paymentmgr.go
```diff
@@ -595,71 +644,237 @@ func (pm *PaymentMgr) maturePendingPayments(height uint32) (map[uint32][]*Paymen
...
+	}
+
+	balanceReq := &walletrpc.BalanceRequest{
```
jholdstock, line 814: Would like to see a comment here, something like 

```
// Balance request not technically required here, just used so we can provide a helpful error message. 
```

pool/paymentmgr.go
```diff
@@ -595,71 +644,237 @@ func (pm *PaymentMgr) maturePendingPayments(height uint32) (map[uint32][]*Paymen
...
+	if err != nil {
+		return fmt.Errorf("unable to publish transaction: %v", err)
+	}
+
+	txid, err := chainhash.NewHash(pubTxResp.TransactionHash)
+	if err != nil {
+		return err
+	}
+
+	log.Tracef("paid a total of %v in tx %s, including %v in pool fees.",
```
jholdstock, line 857: I would promote this from Trace to Debug, possibly even Info. It's really useful information and wont be logged too frequently.

pool/paymentmgr_test.go
```diff
@@ -872,62 +691,50 @@ func testPaymentMgr(t *testing.T, db *bolt.DB) {
 		t.Fatalf("unable to generate payments: %v", err)
 	}
 
-	// Ensure only bundles for account x and fees were created.
-	bundles, err = mgr.fetchEligiblePaymentBundles(paymentMaturity)
+	// Ensure payments for  for account x, y and fees were created.
```
jholdstock, line 694:
```suggestion
	// Ensure payments for account x, y and fees were created.
```

pool/paymentmgr.go
dnldd: The purpose of that was to group payments by the height they were created on, I'll be reworking that actually. needs to be keyed by the coinbase they are paying from. That has more utility.

pool/paymentmgr.go
dnldd: Good catch, will have to think about that some more. What's coming to mind immediately is to split the fees based on the percentages of the amounts being paid out compared to the coinbase total minus the pool fees.

pool/paymentmgr.go
dnldd: I disagree because I see the loop terminating in each case I can think of. `height` remains static as the loop runs, only `walletHeight` is getting updated in the loop. If the connection to the wallet is lost at some point  `WalletBestHeight()` will error and break the loop. If there is an edge case I'm not thinking of let me know.

pool/paymentmgr.go
dnldd: I'd say it is required there, the spendable prev outs from the wallet cannot be less than the total outputs paying from those prev outs.

pool/paymentmgr.go
```diff
@@ -277,33 +289,142 @@ func (pm *PaymentMgr) loadLastPaymentCreatedOn(tx *bolt.Tx) error {
...
+		if tally[account].Cmp(ZeroRat) == 0 {
+			return nil, MakeError(ErrDivideByZero, "division by zero", nil)
+		}
+		accPercent := new(big.Rat).Quo(shareCount, totalShares)
+		percentages[account] = accPercent
+	}
+	return percentages, nil
+}
+
+// PPSEligibleShares fetches all shares within the provided inclusive bounds.
```
jholdstock, line 321: All shares are returned if `min` is `nil`. We should either mention this in the comment, or create a new func which is specifically for returning all shares and takes no params. 

Personally I think the code would be cleaner with the second option.

pool/paymentmgr_test.go
```diff
@@ -98,11 +187,168 @@ func testPaymentMgr(t *testing.T, db *bolt.DB) {
...
+		t.Fatalf("emptyBucket error: %v", err)
+	}
+
+	// Create a share below the minimum exclusive PPLNS time for account x.
+	err = persistShare(db, xID, weight, belowMinimumTime)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Create a share below the minimum exxcuive PPLNS time for account y.
```
jholdstock, line 287:
```suggestion
	// Create a share below the minimum exclusive PPLNS time for account y.
```

pool/paymentmgr_test.go
```diff
@@ -98,11 +187,168 @@ func testPaymentMgr(t *testing.T, db *bolt.DB) {
...
+	}
+
+	// Ensure the returned number of shates is as expected.
+	if len(shares) != expectedShareCount {
+		t.Fatalf("PPLNS error: expected %v eligible PPLNS shares, got %v",
+			expectedShareCount, len(shares))
+	}
+
+	// Ensure account x and account y both have shares returned.
+	if forAccX == 0 || forAccY == 0 {
```
jholdstock, line 329: `forAccX` and `forAccY` have not been recalculated. This is still using the values from the previous test

pool/paymentmgr_test.go
dnldd: resolved in upcoming commit.

pool/paymentmgr.go
dnldd: resolved in upcoming commit.

pool/paymentmgr.go
dnldd: resolved in upcoming commit.

pool/paymentmgr_test.go
dnldd: resolved in upcoming commit.

pool/paymentmgr.go
dnldd: resolved in upcoming commit.

pool/difficulty_test.go
```diff
@@ -8,7 +8,45 @@ import (
...
+		},
+		{
+			new(big.Int).SetInt64(1.2e12),
+			new(big.Int).SetInt64(10),
+			"471159217274235821222212666864778270887329039166831722165761653678014464/48828125",
+		},
+	}
+
+	for _, test := range poolTgts {
+		target, _, err := calculatePoolTarget(chaincfg.MainNetParams(),
```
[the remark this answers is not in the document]
dnldd: Good catch, this led me to a clamping bug, fixed now. :+1:

pool/paymentmgr.go
dnldd: I think the first option is better here considering the call site, don't see how the code will be cleaner since it'd be calling `PPSEligibleShares` internally and passing nil to it.

pool/paymentmgr.go
dnldd: used the block hash as the key to make it easier to prune payments sourcing from orphaned blocks.

pool/paymentmgr_test.go
```diff
@@ -1,45 +1,504 @@
...
+		t.Fatalf("emptyBucket error: %v", err)
+	}
+
+	// Create a share below the minimum exclusive PPLNS time for account x.
+	err = persistShare(db, xID, weight, belowMinimumTime)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Create a share below the minimum excluive PPLNS time for account y.
```
jholdstock, line 287:
```suggestion
	// Create a share below the minimum exclusive PPLNS time for account y.
```

pool/payment.go
```diff
@@ -119,48 +133,16 @@ func (pmt *Payment) Update(db *bolt.DB) error {
 	return pmt.Create(db)
```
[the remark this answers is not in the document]
dnldd: Necessary to separate them here to keep a uniform interface for when different databases are supported.

pool/payment.go
```diff
@@ -119,48 +133,16 @@ func (pmt *Payment) Update(db *bolt.DB) error {
...
-		bundle.Payments[idx].PaidOnHeight = height
-	}
-}
-
-// ArchivePayments removes all payments included in the payment bundle from the
-// payment bucket and archives them.
-func (bundle *PaymentBundle) ArchivePayments(db *bolt.DB) error {
-	err := db.Update(func(tx *bolt.Tx) error {
+// Archive removes the associated payment from active payments and archives it.
+func (pmt *Payment) Archive(db *bolt.DB) error {
```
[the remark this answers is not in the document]
dnldd: disagree here, the pattern you've noticed isn't consistent across call sites. The most important call site which is in `paymentmgr.go` just sets paid on height and archives, what you've described is associated with tests. Also making `Archive` set paid on height gives it multiple tasks than just archiving. Doing that and still calling it `Archive` would be wrong, Opting to keep it as is.

pool/paymentmgr.go
```diff
@@ -614,6 +614,42 @@ func (pm *PaymentMgr) pendingPaymentsAtHeight(height uint32) ([]*Payment, error)
...
+		cursor := bkt.Cursor()
+		for k, v := cursor.First(); k != nil; k, v = cursor.Next() {
+			var payment Payment
+			err := json.Unmarshal(v, &payment)
+			if err != nil {
+				return err
+			}
+
+			if payment.PaidOnHeight == 0 {
+				if payment.Source == nil {
```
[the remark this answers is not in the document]
dnldd: Good catch, that should be removed, in the db update that source is set a zero source, which is a source a coinbase and blockhash being empty strings. The source will not be nil as a result.

pool/paymentmgr.go
```diff
@@ -743,6 +746,32 @@ func (pm *PaymentMgr) payDividends(height uint32) error {
...
+	for key := range pmts {
+		blockHash, err := chainhash.NewHashFromStr(key)
+		if err != nil {
+			return err
+		}
+
+		_, err = txCreator.GetBlock(blockHash)
+		if err != nil {
+			if rpcErr, ok := err.(*dcrjson.RPCError); ok {
+				if rpcErr.Code != dcrjson.ErrRPCBlockNotFound {
```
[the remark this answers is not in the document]
dnldd: Good catch, need to lookup the number of confirmations here to assert if the block is indeed an orphan.

pool/paymentmgr.go
```diff
@@ -834,21 +863,31 @@ func (pm *PaymentMgr) payDividends(height uint32) error {
 		outSizes = append(outSizes, txsizes.P2PKHOutputSize)
 	}
 
-	estSize := txsizes.EstimateSerializeSizeFromScriptSizes(inSizes, outSizes,
-		txsizes.P2PKHOutputSize)
+	estSize := txsizes.EstimateSerializeSizeFromScriptSizes(inSizes, outSizes, 0)
```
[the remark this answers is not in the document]
dnldd: Yes, all output values not consumed by payouts are fees, no change will be generated by the tx.

Issue #208, opened 2020-04-30
Begin restructuring to allow optional DB


githubsands:
@dnldd and I agreed on restructuring dcrpool to easily allow postgres as a database option.  This issue serves to document this discussion.  The goal here is to provide a common interface to use for both `boltdb` and `postgres` databases. 

```
dcrpool/pool/
             /errors
             /db/
                  interface.go
                  /postgres
                  /bolt
 ```

**Comment by jholdstock** [comment]
@dnldd  mentioned something to me whilst reviewing #207. 

We have various entities which will need to be stored in the database, for example, shares and payments. These entities already exist as `shares.go` and `payments.go`. These files contain some oddities, for example `share.go` has methods `Update` and `Delete` which do nothing but return a "Not supported error" , `payment.go` has an `Update` method which does nothing but delegate to the `Create` method. I believe the idea is that each entity which is being stored in the database will have a common interface, which should make this task (#208) easier.

I disagree that these entities should require a common interface like this, and in fact they should not really contain methods like `Create` or `Update` at all . If the database is the thing which we are trying to abstract, it is the database which should have an interface, something like the following

```
type database interface {
    createShare(share) err
    retrieveShare(share) (share, err)
    deleteShare(share) (share, err)
    ... etc
}
```
I believe this is the pattern @githubsands describes above.

With this pattern, there is no need for "Not supported errors" or dummy methods. Instead, an unsupported or not required function would simply not exist in the database interface. The entities do not need to confirm to an interface, and they remain unaware of the database they are being stored in. The entities are simplified to focus only on their own fields and business logic.

dnldd:
#282 closes this.

PR #209, opened 2020-05-01
[WIP] multi: lay the foundations to allow postgres

githubsands:
PR for #208 
Commit 1:
* creates error package to prevent cyclic imports
* fix tests and other imports as a result of the new erorr package


jholdstock:
@dnldd - Most of the work in this PR seems to be redundant because the code it was built upon has changed so much. Also haven't seen @githubsands in a while, so perhaps we should close this PR.

dnldd:
agreed, closing.

PR #213, opened 2020-05-12
config: Only allow pool fee between 0.2% and 5%

### LINKED ISSUE #210: Handling for transaction fee > pool fee
With the new payment code in #207, the pool is going to cover reward payment transaction fees using its own portion of the mining reward. The amount of mining reward claimed by the pool is configurable, any value above zero is allowed. This enables a scenario which is not currently covered - the pool fee is too small to cover the transaction fees. We need special handling for this case because with the current code, no reward payments will ever be sent

jholdstock:
Closes #210

### DIFF
--- a/config.go
+++ b/config.go
@@ -565,7 +565,7 @@ func loadConfig() (*config, []string, error) {
 
 		// Ensure pool fee is valid.
 		if cfg.PoolFee < 0.002 || cfg.PoolFee > 0.05 {
-			str := "%s: poolfee should be between 0.002 and 0.05"
+			str := "%s: poolfee should be between 0.002 (0.2%) and 0.05 (5%)"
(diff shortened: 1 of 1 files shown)


config.go
```diff
@@ -568,8 +568,8 @@ func loadConfig() (*config, []string, error) {
 		}
 
 		// Ensure pool fee is valid.
-		if cfg.PoolFee < 0 || cfg.PoolFee > 1 {
-			str := "%s: poolfee should be between 0 and 1"
+		if cfg.PoolFee < 0.002 || cfg.PoolFee > 0.05 {
+			str := "%s: poolfee should be between 0.002 and 0.05"
```
dnldd, line 572: Please include the corresponding percentage values of the bounds in the error messages as well.

PR #215, opened 2020-05-12
miner: Print value of string, not pointer.

jholdstock:


jholdstock:
Closing this - fix is not appropriate as it can cause panics if `sErr.Traceback` is nil. Will leave with @dnldd to fix as discussed on Matrix

PR #216, opened 2020-05-13
Fix harness to not expose profile

degeri:
The ":" while defining port made the profile service accessible on all interfaces. Now it only contains port number hence it defaults to localhost.


harness.sh
```diff
@@ -105,7 +105,7 @@ lastnperiod=${LAST_N_PERIOD}
 adminpass=${ADMIN_PASS}
 guidir=${GUI_DIR}
 designation=${SESSION}
-profile=:6060
+profile=6060
```
jholdstock, line 108: Also needs updating on line 72

PR #217, opened 2020-05-13
Add dcrctl install instructions for harness

degeri:
harness needs dcrctl to work. Since dcrctl has it own repo now it needs to be installed separately from dcrd.


degeri:
So the configuration is covered here  https://github.com/decred/dcrpool#installing-and-updating and here https://github.com/decred/dcrpool#configuration

Issue #218, opened 2020-05-13
Review usage of multiple pool fee addresses


jholdstock:
Config allows multiple addresses to be provided for collecting pool fees. I had assumed that fees would be split equally among all provided addresses, however it seems that each time a payment is processed, a single fee address is selected **at random**.

https://github.com/decred/dcrpool/blob/2b8ba4522724d2c3ef7eccf415fbc4a91184ff9a/pool/paymentmgr.go#L748-L749

I am assuming this is a temporary behaviour and something which needs to be fixed. If this is a deliberate feature then it needs to be clearly documented with proper justification.

dnldd:
Will get this done soon.

dnldd:
Resolved by #225

PR #225, opened 2020-05-14
multi: utilize tx confimation notifications.

### LINKED ISSUE #222: Utilize wallet tx confirmation notifications.
Currently coinbases are being monitored via the `getbalance` rpc, the wallet sync height and waiting a few seconds to avoid spending the coinbase before the wallet is aware of it. This is inefficient and error prone. It'd be ideal to monitor coinbase txs via wallet notifications instead because a received notification confirming outputs of a tx are mature and spendable implicitly also confirm the wallet is aware of the tx as well. In contrast, `gettxout` does not confirm the latter. 

S/O to @jholdstock  for the suggestion.

dnldd:
This uses tx confirmation notifications from the wallet to confirm the coinbases sourced by payments are spendable before attempting to spend them. The readme now  better explains why multiple pool fee address and multiple mining addresses are important to set in production.

Resolves #222 and #218.

### DIFF
--- a/pool/paymentmgr.go
+++ b/pool/paymentmgr.go
@@ -748,7 +748,7 @@ func (pm *PaymentMgr) payDividends(ctx context.Context, height uint32) error {
 	// The fee address is being picked at random from the set of pool fee
 	// addresses to make it difficult for third-parties wanting to track
 	// pool fees collected by the pool and ultimately determine the
-	// cummulative value accrued by pool operators.
+	// cumulative value accrued by pool operators.
 	feeAddr := pm.cfg.PoolFeeAddrs[rand.Intn(len(pm.cfg.PoolFeeAddrs))]
 
 	// Create the payout transaction.
--- a/pool/paymentmgr.go
+++ b/pool/paymentmgr.go
@@ -747,8 +747,13 @@ func (pm *PaymentMgr) payDividends(ctx context.Context, height uint32) error {
 		return nil
 	}
 
-	// Create the payout transaction.
+	// The fee address is being picked at random from the set of pool fee
+	// addresses to make it difficult for third-parties wanting to track
+	// pool fees collected by the pool and ultimately determine the
+	// cummulative value accrued by pool operators.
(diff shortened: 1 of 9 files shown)


pool/paymentmgr.go
```diff
@@ -904,53 +916,64 @@ func (pm *PaymentMgr) payDividends(ctx context.Context, height uint32) error {
...
+
+				// Remove spendable coinbases from the tx hash set. All
+				// coinbases are spendable when the tx hash set is empty.
+				delete(txHashes, cbHash.String())
+			}
 		}
 
-		if walletHeight >= height {
-			time.Sleep(time.Second * 3)
+		if len(txHashes) == 0 {
```
jholdstock, line 959: Should we close notification stream here?

pool/paymentmgr.go
```diff
@@ -904,53 +916,64 @@ func (pm *PaymentMgr) payDividends(ctx context.Context, height uint32) error {
...
+		maxSpendableHeight = height
 	}
 
-	txB, err := tx.Bytes()
+	hashes := make([]*chainhash.Hash, 0, len(txHashes))
+	for _, hash := range txHashes {
+		hashes = append(hashes, hash)
+	}
+
+	notifSource, err := pm.cfg.GetTxConfNotifications(hashes,
```
jholdstock, line 930: This seems to be creating a new notification stream every time payDividends is called. Are the old streams properly closed?

pool/paymentmgr.go
dnldd: Nope, the stream is terminated at `maxSpendableHeight`.

pool/paymentmgr.go
dnldd: No it's not, each requested stream terminates at the provided `StopAfter` height. `GetTxNotifications` is only returning the function to read the stream with after the confirmation notification request is sent. This function is the `Recv` of the confirmation notification client.

jholdstock:
Tested thoroughly on testnet with 6 mining clients, using both PPS and PPLNS. Performed start/stop testing. Everything is looking good :+1:

PR #238, opened 2020-06-18
Use separate directories for each Decred network

jholdstock:


config.go
```diff
@@ -494,8 +494,26 @@ func loadConfig() (*config, []string, error) {
...
+		return nil, nil, fmt.Errorf("unknown network provided %v",
+			cfg.ActiveNet)
+	}
+
 	cfg.DataDir = cleanAndExpandPath(cfg.DataDir)
 	cfg.LogDir = cleanAndExpandPath(cfg.LogDir)
+
+	// Use separate data and log directories for each Decred network.
+	cfg.DataDir = filepath.Join(cfg.DataDir, cfg.net.Name)
+	cfg.LogDir = filepath.Join(cfg.LogDir, cfg.net.Name)
```
dnldd, line 515: how about passing the results of path joining to `cleanAndExpanPath`,  for `cfg.DataDir` it'd be: `cfg.DataDir = cleanAndExpandPath(filepath.Join(cfg.DataDir, cfg.net.Name))`

jholdstock: Updated

PR #243, opened 2020-08-17
Various harness updates

jholdstock:
- Add more detailed docs to the harness, remove details from project README
- Rename some vars in the harness to be more accurate:
  - `NODES_ROOT` becomes `HARNESS_ROOT`
  - `SESSION` becomes `TMUX_SESSION`
- tmux session is now `dcrpool-harness` rather than simply `harness`
- Create harness directory in `/tmp` rather than user home dir
- Create wallets with stdin rather than using tmux send-keys & sleep (decred/dcrwallet#1807)

Also including the update to go 1.15

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -6,7 +6,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: [1.13, 1.14]
+        go: [1.14, 1.15]
     steps:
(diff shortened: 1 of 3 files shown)


jholdstock:
Pushed one more commit which updates addresses to SLIP0044 (to accommodate decred/dcrwallet#1818)

dnldd (approved):
Nicely done!

PR #244, opened 2020-08-26
harness.sh: utilize tmux's wait-for.

dnldd:
**Depends on #243**

This updates the harness to remove unneeded sleeps and utilize tmux's wait-for where applicable.


jholdstock:
#243

dnldd:
Just noticed the wallet is generating different addresses with the same seed after updating the wallet. Taking a closer look.

jholdstock (commented):
Looks good on the whole, just unsure why you have added `--addrindex` to the dcrd nodes?

We don't recommend using --addrindex on testnet/mainnet deployments, so I would avoid using them on the simnet harness in case it masks some kind of issue in future.

dnldd:
I wanted to have that index available to use, I've needed it in a couple of dcrd related tests. I agree it should not be enabled by default, will remove it.

jholdstock (approved):

PR #245, opened 2020-09-07
pool: update error types.

dnldd:
This updates the error types to leverage go 1.13 errors.Is/As functionality, conforming to the updated error infrastructure detailed [here](https://github.com/decred/dcrd/issues/2181).

PR #247, opened 2020-09-17
Use existing account if one exists

jholdstock:
There is currently an issue introduced by #245 where a disconnected miner is not able to reconnect because of an "account already exists" error:

```
2020-09-17 10:39:07.242 [DBG] POOL: Mining client connected. id=a42490b0/cpu, addr=127.0.0.1:36458
2020-09-17 10:39:07.263 [ERR] POOL: Account.Create: account 7f36b2c979067add6ce030176e0bdc75ac65798842f339aed8ad22d6cdda584c already exists
2020-09-17 10:39:07.283 [ERR] POOL: a42490b0/cpu: EOF
```

This PR fixes the issue by introducing and using a new func `CreateOrFetchAccount`. `CreateOrFetchAccount` will retrieve an account if one exists, or create one if not.

The check for address decoding is also moved into `CreateOrFetchAccount`. We only need to check for address validity once before inserting an account into the database, we don't need to do it every single time we just want to create an account ID.

### DIFF
--- a/gui/account.go
+++ b/gui/account.go
@@ -36,11 +36,7 @@ func (ui *GUI) account(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Generate the account id of the provided address.
-	accountID, err := pool.AccountID(address, ui.cfg.ActiveNet)
-	if err != nil {
-		ui.renderIndex(w, r, "Unable to generate account ID for address")
-		return
-	}
+	accountID := pool.AccountID(address)
 
 	if !ui.cfg.AccountExists(accountID) {
 		ui.renderIndex(w, r, "Nothing found for address")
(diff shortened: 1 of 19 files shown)


dnldd:
I think the solution could be a lot simpler. When the client is being created in `handleAuthorizeRequest` the error needs to be filtered there to ignore `ErrValueFound` and proceed to set the account id to the mining client. 

```
err = account.Create(c.cfg.DB)
if err != nil {
	// Do not error if the account already exists.
	if !errors.Is(err, ErrValueFound) {
		sErr := NewStratumError(Unknown, err)
		resp := AuthorizeResponse(*req.ID, false, sErr)
		c.ch <- resp
		return err
	}
}
c.account = account.UUID
c.name = name
```

dnldd (approved):
Tested the changes made, looks good.

Issue #200, opened 2020-04-22
Show pool fees collected in a given time period on admin page.


dnldd:
The admin page should be able to show pool fees gained in given a time period.

PR #249, opened 2020-09-17
gui: Show pool payments.

jholdstock:
Pending and archived payments to the pool are now displayed on the admin page.

First step towards #200

### DIFF
--- a/gui/account.go
+++ b/gui/account.go
@@ -14,14 +14,16 @@ import (
 // accountPageData contains all of the necessary information to render the
 // account template.
 type accountPageData struct {
-	HeaderData       headerData
-	MinedWork        []*minedWork
-	ArchivedPayments []*archivedPayment
-	PendingPayments  []*pendingPayment
-	ConnectedClients []*client
-	AccountID        string
-	Address          string
-	BlockExplorerURL string
+	HeaderData            headerData
+	MinedWork             []*minedWork
+	ArchivedPaymentsTotal string
+	ArchivedPayments      []*archivedPayment
+	PendingPaymentsTotal  string
+	PendingPayments       []*pendingPayment
+	ConnectedClients      []*client
+	AccountID             string
+	Address               string
+	BlockExplorerURL      string
 }
 
 // account is the handler for "GET /account". Renders the account template if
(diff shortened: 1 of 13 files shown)


dnldd:
I think this needs some more work. Since it's an admin page I think we should be showing information that is glanceable: total payments pending most importantly and the total value accrued by pool fees (I need to do some additional work to expose a call for this). The page should also support the usecase where the admin needs to confirm a complaint from a pool user concerning a pending payment. With the user's address or account id the admin should be able to perform a search and see which pending payments are still due the user and which payments have already been processed as well as the blocks the user's mining clients mined on behalf of the pool.

jholdstock:
I will include some additional changes to also display total fees accrued by the pool (and probably also add this to the user payments too). I don't think we need a new call exposing this from the backend because this can be calculated by the front-end using data already in the cache.

> The page should also support the usecase where the admin needs to confirm a complaint from a pool user concerning a pending payment. With the user's address or account id the admin should be able to perform a search and see which pending payments are still due the user and which payments have already been processed as well as the blocks the user's mining clients mined on behalf of the pool.

All of this information is already displayed to the user when searching for their address on the home page. I don't believe there is any additional information stored in the database which is not being displayed. 

The only thing missing is the ability to search by account ID, but is that something which would actually be useful?

jholdstock:
I've got some local changes for displaying pending/archived payment totals, but I haven't been table to test locally yet because of the panic described in #251

jholdstock:
Rebased and pushed a new commit which displays pending/archived payment totals

dnldd:
This looks good, I have expanded on what needs further refinement or is missing from the admin section. I think we can proceed with merging this while the others are tackled in separate PRs.

- the layout of the the admin page currently isn't ideal, it's has the information needed but the layout and overall presentation could be bettered. We'll need to get in touch with the designer to get an updated layout to work with but in in the mean time we can proceed with adding the needed info/features.

- currently the pending/archived payments need to be scrolled to to be seen, ideally all glanceable info on the admin page should 
be in the header and always viewable on the page. We'll need to provide this info to the designer to consider when updating the 
admin layout.

- currently the admin page does not allow the admin to lookup a pool account to either cross-check or confirm an issue reported 
by a user. I think this functionality needs to be provided. We'll either have to dedicate a page to it or find a clean way to add it to 
the current page. I'd say let's get this done before we get in touch with the designer to rework the layout.

dnldd (approved):

PR #253, opened 2020-10-06
Remove CreatedOn timestamp from payments

jholdstock:
- List "Estimated Payment Height" for pending payments instead of CreatedOn.
- Display payment tables side-by-side on large screens. Maintain vertical stacking on smaller screens + mobile.
- Fix payment totals alignment issues

Screenshot from 2020-10-06 11-36-25

### DIFF
--- a/gui/cache.go
+++ b/gui/cache.go
@@ -45,10 +45,10 @@ type rewardQuota struct {
 // pendingPayment represents an unpaid reward payment. It is json annotated so
 // it can easily be encoded and sent over a websocket or pagination request.
 type pendingPayment struct {
-	WorkHeight    string `json:"workheight"`
-	WorkHeightURL string `json:"workheighturl"`
-	Amount        string `json:"amount"`
-	CreatedOn     string `json:"createdon"`
+	WorkHeight             string `json:"workheight"`
+	WorkHeightURL          string `json:"workheighturl"`
+	Amount                 string `json:"amount"`
+	EstimatedPaymentHeight string `json:"estimatedpaymentheight"`
 }
 
 // archivedPayment represents a paid reward payment. It is json annotated so it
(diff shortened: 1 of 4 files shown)


gui/cache.go
```diff
@@ -45,10 +45,10 @@ type rewardQuota struct {
 // pendingPayment represents an unpaid reward payment. It is json annotated so
 // it can easily be encoded and sent over a websocket or pagination request.
 type pendingPayment struct {
-	WorkHeight    string `json:"workheight"`
-	WorkHeightURL string `json:"workheighturl"`
-	Amount        string `json:"amount"`
-	CreatedOn     string `json:"createdon"`
+	WorkHeight             string `json:"workheight"`
```
dnldd, line 48: Just noticing this, why aren't `WorkHeight`, `Amount` and `EstimatedPaymentHeight` their default types here? I reckon the templates could have formatting needed to cast the types when needed.

gui/cache.go
```diff
@@ -290,10 +289,10 @@ func (c *Cache) updatePayments(pendingPmts []*pool.Payment, archivedPmts []*pool
...
 		pendingPayments[accountID] = append(pendingPayments[accountID],
 			&pendingPayment{
-				WorkHeight:    fmt.Sprint(p.Height),
-				WorkHeightURL: blockURL(c.blockExplorerURL, p.Height),
-				Amount:        amount(p.Amount),
-				CreatedOn:     formatUnixTime(p.CreatedOn),
+				WorkHeight:             fmt.Sprint(p.Height),
+				WorkHeightURL:          blockURL(c.blockExplorerURL, p.Height),
+				Amount:                 amount(p.Amount),
+				EstimatedPaymentHeight: fmt.Sprint(p.EstimatedMaturity + 1),
```
dnldd, line 295: The string conversions here will not be needed if the types are their default types.

dnldd (changes requested):

gui/cache.go
jholdstock: If we rely upon the templates to execute the formatting logic then it will be performed on every single field, every single time a page is loaded.

The formatting is currently executed before the data is inserted into the cache, which means it is only performed once and never duplicated unnecessarily.

dnldd: noted, thanks.

dnldd (approved):

PR #256, opened 2020-10-09
Handle database IDs as strings where possible.

jholdstock:
Regular conversion of database IDs between `string` and `[]byte` is necessary because bbolt expects all keys and values to be byte arrays. Because this is specific to bbolt and we want to generalise the code to work with other database implementations, the majority of our code should handle IDs as `string` and only convert to `[]byte` as close to the bbolt code as possible.

This has a nice side-effect of removing a large amount of inline type conversion.

Part of #257


dnldd (approved):
Looks good. Would be great to have a issue for postgres support where we could link this PR and all upcoming PRs on this for tracking purposes.

Issue #257, opened 2020-10-10
Postgres


jholdstock:
This issue will serve as an umbrella issue for all of the changes needed to support Postgres as a database option.

The intention is to add new config items which will allow an administrator to select between using the current bbolt database, or an external postgres database.

This will require separating the current database code from the the codebase, making it generic enough to create an abstraction layer, re-implementing the current db code in Postgres, and adding the new config items which will allow switching between the two modes.

jholdstock:
The above PRs extract the majority of the database code from the rest of the code. I will soon push a PR which abstracts the database behind an interface which looks something like the following:

```
type database interface {
    createShare(share) err
    retrieveShare(share) (share, err)
    deleteShare(share) (share, err)
    ... etc
}
```

We need to start considering how to handle database backups. dcrpool currently:
- writes a database backup on shutdown
- allows downloading a database backup through the web UI

Another consideration is the behaviour when dcrpool is changed from solo mode to pool mode (or vice versa). Currently we make a backup of the database file and start a new empty database in the new mode. Is this desirable (or even feasible) with postgres?

dnldd:
I think we should maintain writing a database backup as well on shutdown as downloading a backup via the web UI for the bolt db driver, the postgres db driver will not support these functionality. Backing up for postgres will require a separate db instance synchronizing to the live instance. We can outline how to set this up in the README, I think that's as far as we go for postgres regarding backup.

jholdstock:
I've hit a little bit of a roadblock because of how heavily the tests depend on the database implementation. To provide an example, when the payment manager is tested, we pass it a real database implementation and then we verify the paymentmgr behaviour by checking the contents of the database. A good example is here https://github.com/decred/dcrpool/blob/11b44c239076484027d546bb795d50a68a4f9fac/pool/paymentmgr_test.go#L192-L212

This single test is verifying both the paymentmgr behaviour, and the database access code.

In a perfect world we would pass the payment manager a mock database and validate its behaviour against the mock, rather than using a real db.
We would then have a separate test for database access code.

And in future, when we add postgres support, we would add a new test for database access via postgres, and leave the payment mgr test untouched

jholdstock:
Encountered an issue with how shares are encoded and stored in boltDB.

Share created on time is not stored as a discrete value in boltDB. Instead, converted from an integer value into bytes, hex encoded, and stored as the first 16 characters of the share ID.

[...]

PR #258, opened 2020-10-10
Extract db code from chainstate.go

jholdstock:
Part of #257

### DIFF
--- a/pool/acceptedwork.go
+++ b/pool/acceptedwork.go
@@ -191,9 +191,9 @@ func ListMinedWork(db *bolt.DB) ([]*AcceptedWork, error) {
 	return minedWork, nil
 }
 
-// FetchUnconfirmedWork returns all work which is not confirmed as mined with
+// fetchUnconfirmedWork returns all work which is not confirmed as mined with
 // height less than the provided height.
-func FetchUnconfirmedWork(db *bolt.DB, height uint32) ([]*AcceptedWork, error) {
+func fetchUnconfirmedWork(db *bolt.DB, height uint32) ([]*AcceptedWork, error) {
 	toReturn := make([]*AcceptedWork, 0)
 	err := db.View(func(tx *bolt.Tx) error {
 		bkt, err := fetchBucket(tx, workBkt)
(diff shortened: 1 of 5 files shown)


pool/acceptedwork.go
```diff
@@ -208,3 +208,42 @@ func ListMinedWork(db *bolt.DB) ([]*AcceptedWork, error) {
 	}
 	return minedWork, nil
 }
+
+// FetchUnconfirmedWork returns all work which is not confirmed as mined with
+// height less than the provided height.
+func FetchUnconfirmedWork(db *bolt.DB, height uint32) ([]*AcceptedWork, error) {
```
dnldd, line 214: Let's unexport the new helpers, I do not see any other call sites for them outside of the pool package currently.

pool/chainstate.go
```diff
@@ -235,41 +177,6 @@ func (cs *ChainState) prunePayments(ctx context.Context, height uint32) error {
 	return nil
 }
 
-// pruneJobs removes all jobs with heights less than the provided height.
-func (cs *ChainState) pruneJobs(height uint32) error {
```
dnldd, line 239: I think the chainstate should maintain the pruning functions like you did for `pruneAcceptedWork` and extracting a helper fetching function out of that instead.

pool/payment.go
```diff
@@ -162,3 +162,40 @@ func (pmt *Payment) Archive(db *bolt.DB) error {
 		return nil
 	})
 }
+
+// fetchOrphanedPayments returns all payments sourcing from orphaned blocks at
+// the provided height.
+func fetchOrphanedPayments(db *bolt.DB, height uint32) ([]*Payment, error) {
```
dnldd, line 168: Would suggest this be renamed to `pendingPaymentsAtHeight` since at this point the payments returned have not yet been confirmed as being sourced from orphaned blocks.

dnldd (changes requested):

pool/payment.go
jholdstock: Done

dnldd:
On further discussion we've agreed to make the needed api changes since the functions that are currently being made standalone will end up as part of a db driver in the long run and also eliminate the amount or wrapping needed to maintain the current api. Do proceed.

dnldd (approved):

PR #259, opened 2020-10-10
Extract db code from paymentmgr.go

jholdstock:
Part of #257

### DIFF
--- a/pool/paymentmgr.go
+++ b/pool/paymentmgr.go
@@ -394,12 +394,18 @@ func (pm *PaymentMgr) sharePercentages(shares []*Share) (map[string]*big.Rat, er
 	return percentages, nil
 }
 
+// PPSEligibleShares fetches all shares created before or at the provided
+// time.
+func (pm *PaymentMgr) PPSEligibleShares(max []byte) ([]*Share, error) {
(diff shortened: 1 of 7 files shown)


pool/paymentmgr.go
```diff
@@ -395,51 +394,12 @@ func (pm *PaymentMgr) sharePercentages(shares []*Share) (map[string]*big.Rat, er
 	return percentages, nil
 }
 
-// PPSEligibleShares fetches all shares created before or at the provided
-// time.
-func (pm *PaymentMgr) PPSEligibleShares(max []byte) ([]*Share, error) {
```
dnldd, line 400: Let's give the `paymentmgr` as `chainstate` the same treatment: maintain the current api and extract unexported db helpers where needed.

dnldd (changes requested):

pool/paymentmgr.go
jholdstock: Done

dnldd: On further discussion we've agreed to make the needed api changes since the functions that are currently being made standalone will end up as part of a db driver in the long run and also eliminate the amount or wrapping needed to maintain the current api. Do proceed.

dnldd (approved):

PR #261, opened 2020-10-12
Extract more db code from paymentmgr.go

jholdstock:
Part of #257

### DIFF
--- a/pool/paymentmgr.go
+++ b/pool/paymentmgr.go
@@ -204,19 +204,21 @@ func (pm *PaymentMgr) fetchLastPaymentHeight() uint32 {
 	return atomic.LoadUint32(&pm.lastPaymentHeight)
 }
 
-// persistLastPaymentHeight saves the last payment height to the db.
-func (pm *PaymentMgr) persistLastPaymentHeight() error {
+// persistLastPaymentInfo saves the last payment height and time to the db.
+func (pm *PaymentMgr) persistLastPaymentInfo() error {
 	height := pm.fetchLastPaymentHeight()
-	return persistLastPaymentHeight(pm.cfg.DB, height)
+	paidOn := pm.fetchLastPaymentPaidOn()
+	return persistLastPaymentInfo(pm.cfg.DB, height, int64(paidOn))
 }
 
-// loadLastPaymentHeight fetches the last payment height from the db.
-func (pm *PaymentMgr) loadLastPaymentHeight() error {
-	height, err := loadLastPaymentHeight(pm.cfg.DB)
+// loadLastPaymentInfo fetches the last payment height and time from the db.
+func (pm *PaymentMgr) loadLastPaymentInfo() error {
+	height, paidOn, err := loadLastPaymentInfo(pm.cfg.DB)
 	if err != nil {
 		return err
 	}
 	pm.setLastPaymentHeight(height)
+	pm.setLastPaymentPaidOn(paidOn)
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


pool/paymentmgr.go
```diff
@@ -207,37 +205,18 @@ func (pm *PaymentMgr) fetchLastPaymentHeight() uint32 {
...
 
 // persistLastPaymentHeight saves the last payment height to the db.
-func (pm *PaymentMgr) persistLastPaymentHeight(tx *bolt.Tx) error {
-	funcName := "persistLastPaymentHeight"
-	pbkt, err := fetchPoolBucket(tx)
-	if err != nil {
-		return err
-	}
+func (pm *PaymentMgr) persistLastPaymentHeight() error {
 	height := atomic.LoadUint32(&pm.lastPaymentHeight)
```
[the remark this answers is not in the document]
dnldd: No, the payment manager keeps track of the payment height after loading it initially, it is the tracked value that should be persisted.

pool/paymentmgr.go
```diff
@@ -252,51 +231,13 @@ func (pm *PaymentMgr) fetchLastPaymentPaidOn() uint64 {
...
-	err = pbkt.Put(lastPaymentPaidOn,
-		nanoToBigEndianBytes(int64(pm.lastPaymentPaidOn)))
-	if err != nil {
-		desc := fmt.Sprintf("%s: unable to persist last payment "+
-			"paid on time: %v", funcName, err)
-		return dbError(ErrPersistEntry, desc)
-	}
-	return nil
+func (pm *PaymentMgr) persistLastPaymentPaidOn() error {
+	return persistLastPaymentPaidOn(pm.cfg.DB, int64(pm.lastPaymentPaidOn))
```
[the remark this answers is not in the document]
dnldd: same here

pool/paymentmgr.go
```diff
@@ -538,15 +455,12 @@ func (pm *PaymentMgr) payPerShare(source *PaymentSource, amt dcrutil.Amount, hei
...
-		}
-		return pm.pruneShares(tx, workCreatedOn)
-	})
-	return err
+	// Update the last payment created on time and prune invalidated shares.
+	err = pm.persistLastPaymentCreatedOn()
+	if err != nil {
+		return err
+	}
+	return pruneShares(pm.cfg.DB, workCreatedOn)
```
dnldd, line 463: This ends up being two db calls instead of one, I think you should be utilizing the dbtx  there to group db calls, as was done prior.

pool/paymentmgr.go
```diff
@@ -569,16 +483,13 @@ func (pm *PaymentMgr) payPerLastNShares(source *PaymentSource, amt dcrutil.Amoun
...
-		return pm.pruneShares(tx, minNano)
-	})
-	return err
+	// Update the last payment created on time and prune invalidated shares.
+	err = pm.persistLastPaymentCreatedOn()
+	if err != nil {
+		return err
+	}
+	minNano := time.Now().Add(-pm.cfg.LastNPeriod).UnixNano()
+	return pruneShares(pm.cfg.DB, minNano)
```
dnldd, line 492: Same here, this is now more expensive, two db calls instead of one.

pool/paymentmgr.go
```diff
@@ -1044,15 +955,14 @@ func (pm *PaymentMgr) payDividends(ctx context.Context, height uint32, treasuryA
...
-		return pm.persistLastPaymentPaidOn(tx)
-	})
+	pm.setLastPaymentHeight(height)
+	err = pm.persistLastPaymentHeight()
+	if err != nil {
+		return err
+	}
+
+	pm.setLastPaymentPaidOn(uint64(time.Now().UnixNano()))
+	err = pm.persistLastPaymentPaidOn()
```
dnldd, line 965: now more expensive, two db calls instead of one.

dnldd:
resolved all issues raised because of plans to combine most of the calls highlighted into single functions, after the db interface work is done. We'll revisit this if needed.

dnldd (approved):

PR #263, opened 2020-10-13
Cache last payment details properly.

jholdstock:
Last payment details (height, paid-on, created-on) were previously being cached in atomic vars in the payment manager. This PR caches them in the cache instead.


gui/cache.go
```diff
@@ -382,11 +382,17 @@ func (c *Cache) getArchivedPayments(first, last int, accountID string) (int, []*
 }
 
 func (c *Cache) getArchivedPaymentsTotal(accountID string) string {
+	c.archivedPaymentsMtx.RLock()
+	defer c.archivedPaymentsMtx.RUnlock()
```
dnldd, line 407: I think the `defer` can be avoided for these.

```go
func (c *Cache) getArchivedPaymentsTotal(accountID string) string {
	c.archivedPaymentsMtx.RLock()
	total := c.archivedPaymentTotals[accountID]
	c.archivedPaymentsMtx.RUnlock()	
	return amount(total)
}

func (c *Cache) getPendingPaymentsTotal(accountID string) string {
	c.pendingPaymentsMtx.RLock()
	total := c.pendingPaymentTotals[accountID]
	c.pendingPaymentsMtx.RUnlock()
	return amount(total)
}
```

gui/cache.go
```diff
@@ -337,6 +343,21 @@ func (c *Cache) updatePayments(pendingPmts []*pool.Payment, archivedPmts []*pool
 	c.archivedPaymentsMtx.Unlock()
 }
 
+func (c *Cache) updateLastPaymentInfo(height uint32, paidOn, createdOn int64) {
+	c.lastPaymentInfoMtx.Lock()
+	defer c.lastPaymentInfoMtx.Unlock()
```
dnldd, line 348: same here, the `defer` can be avoided.

pool/db.go
```diff
@@ -368,10 +368,10 @@ func persistLastPaymentInfo(db *bolt.DB, height uint32, paidOn int64) error {
 	})
 }
 
-func loadLastPaymentInfo(db *bolt.DB) (uint32, uint64, error) {
+func loadLastPaymentInfo(db *bolt.DB) (uint32, int64, error) {
 	funcName := "loadLastPaymentInfo"
 	var height uint32
-	var paidOn uint64
+	var paidOn int64
```
dnldd, line 374: What's the reason for changing the type here and for `loadLastPaymentCreatedOn`? We are expecting any negative values for the time so I'd think maintaining `uint64` would adequate. Thoughts?

gui/cache.go
jholdstock: done

gui/cache.go
jholdstock: done

pool/db.go
dnldd: I noticed the created-on/paid-on times were being cast to int64 so maintaining this avoids the casting. Lets maintain the type changes made here.

dnldd (approved):

PR #264, opened 2020-10-13
Separate DB testing from functional testing

jholdstock:
This PR extracts a lot of tests from `paymentmgr_test.go` and `chainstate_test.go` which were actually testing DB helpers and not the `paymentmgr.go` or `chainstate.go` code.

The fact that these tests were actually testing DB helpers was hidden by the fact that the helpers were "wrapped" in `paymentmgr`/`chainstate` funcs which did nothing.

**Note:** Most of the the testing code has simply been moved without changes, however the two tests in `db_tests.go` have been partially rewritten and are worthy of extra scrutiny.

Part of #257


pool/share.go
```diff
@@ -96,7 +96,7 @@ func ppsEligibleShares(db *bolt.DB, max []byte) ([]*Share, error) {
 				return dbError(ErrDecode, desc)
 			}
 
-			if bytes.Compare(createdOnB, max) <= 0 {
+			if bytes.Compare(createdOnB, nanoToBigEndianBytes(max)) <= 0 {
```
dnldd, line 99: this is going to call `nanoToBigEndianBytes` on `max` for every single comparison. better to do this conversion outside the loop so this is done only once.

pool/share.go
```diff
@@ -135,7 +135,7 @@ func pplnsEligibleShares(db *bolt.DB, min []byte) ([]*Share, error) {
 				return dbError(ErrDecode, desc)
 			}
 
-			if bytes.Compare(createdOnB, min) > 0 {
+			if bytes.Compare(createdOnB, nanoToBigEndianBytes(min)) > 0 {
```
dnldd, line 138: same here `nanoToBigEndianBytes` is going to be called once each comparison when it could be done just one outside the loop.

pool/share.go
jholdstock: nice find, fixed

dnldd (approved):

PR #266, opened 2020-10-15
update wallet grpc credentials setup.

dnldd:
This updates the wallet gprc credentials setup to incoming v1.6.0 requirements.

S/O to @jrick  for the heads up.


jrick:
This is reusing the server's cert/key to also act as a client authentication cert? I suppose this works, but it's strange to me.

jrick (approved):
reads fine now

jholdstock (approved):

PR #267, opened 2020-10-22
Introduce Database interface.

jholdstock:
All required database functionality is now defined in a new Database interface. Almost all of the production code has been updated to only handle this interface, rather than any concrete db implementations. This should enable the bbolt implementation to be switched out for a postgres implementation in future.

A big step for #257

### DIFF
--- a/pool/database.go
+++ b/pool/database.go
@@ -24,16 +24,16 @@ type Database interface {
 	loadLastPaymentCreatedOn() (int64, error)
 
 	// Account
-	FetchAccount(id string) (*Account, error)
-	PersistAccount(acc *Account) error
-	DeleteAccount(id string) error
+	fetchAccount(id string) (*Account, error)
+	persistAccount(acc *Account) error
+	deleteAccount(id string) error
 
 	// Payment
-	FetchPayment(id string) (*Payment, error)
-	PersistPayment(payment *Payment) error
-	UpdatePayment(payment *Payment) error
-	DeletePayment(payment *Payment) error
-	ArchivePayment(payment *Payment) error
+	fetchPayment(id string) (*Payment, error)
+	persistPayment(payment *Payment) error
+	updatePayment(payment *Payment) error
+	deletePayment(payment *Payment) error
+	archivePayment(payment *Payment) error
 	fetchPaymentsAtHeight(height uint32) ([]*Payment, error)
 	fetchPendingPayments() ([]*Payment, error)
 	pendingPaymentsForBlockHash(blockHash string) (uint32, error)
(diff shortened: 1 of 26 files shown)


pool/database.go
```diff
@@ -0,0 +1,67 @@
...
+	persistPoolMode(mode uint32) error
+	fetchCSRFSecret() ([]byte, error)
+	persistCSRFSecret(secret []byte) error
+	persistLastPaymentInfo(height uint32, paidOn int64) error
+	loadLastPaymentInfo() (uint32, int64, error)
+	persistLastPaymentCreatedOn(createdOn int64) error
+	loadLastPaymentCreatedOn() (int64, error)
+
+	// Account
+	FetchAccount(id string) (*Account, error)
```
dnldd, line 27: I suspect almost all the functions required by the interface here could be unexported. Will look into doing that if you decide against it in this PR.

pool/hub.go
```diff
@@ -142,7 +141,7 @@ type HubConfig struct {
 type Hub struct {
 	clients int32 // update atomically.
 
-	db             *bolt.DB
+	db             Database
```
dnldd, line 144: Just noticed a the hub config and the hub all have references to the db, likely oversight on my part. I think we should remove one however: the hub reference I'm thinking.

pool/job.go
```diff
@@ -48,10 +48,10 @@ func NewJob(header string, height uint32) *Job {
 }
 
 // FetchJob fetches the job referenced by the provided id.
-func FetchJob(db *bolt.DB, id string) (*Job, error) {
+func (db *BoltDB) FetchJob(id string) (*Job, error) {
```
dnldd, line 51: Perhaps all the `BoltDB` interface implementations need to go into a separate file. Or do you plan on adding the postgres implementations to the struct files as well?

jholdstock: That's something we need to discuss. We could take either approach, but tbh I think both are kind of ugly :joy: 

I havent moved the funcs in this PR simply because it would make the changeset massive, and the review painful.

pool/hub.go
jholdstock: Personally I think removing the one from config makes more sense. The database is one dependency, the config is another. Each is a saparate thing with distinct responsibilities - the config holds simple values which define how the hub should behave, the database is a persistent storage device.

jholdstock: On second thought, I wont make that change because that will conflict with how the rest of the dcrpool components are setup. They all have the database stored in their config, so I will maintain that for hub as well.

pool/job.go
dnldd: Sure thing, let's resolve this in a separate PR.

dnldd (approved):

PR #269, opened 2020-10-23
Remove mode switching behaviour.

jholdstock:
This remove the behaviour which would automatically empty the database and create a new one when the pool mode was changed in config. The new behaviour is to log an error and prevent the pool from starting up.

The reason for this is the empty/backup behaviour does not map well onto other database implementations such as postgres, and after discussing with dnldd we agreed that its better for an admin to explicitly make the decision to backup and create a new DB, rather than the code doing it.

#257

### DIFF
--- a/pool/boltdb.go
+++ b/pool/boltdb.go
@@ -144,66 +144,8 @@ func (db *BoltDB) backup(backupFileName string) error {
 	})
 }
 
-// purge removes all existing data and recreates the db.
-func purge(db *BoltDB) error {
-	const funcName = "purge"
-	err := db.DB.Update(func(tx *bolt.Tx) error {
-		pbkt := tx.Bucket(poolBkt)
-		if pbkt == nil {
-			desc := fmt.Sprintf("%s: bucket %s not found",
-				funcName, string(poolBkt))
-			return dbError(ErrBucketNotFound, desc)
-		}
-		err := pbkt.DeleteBucket(accountBkt)
-		if err != nil {
-			return err
-		}
-		err = pbkt.DeleteBucket(shareBkt)
-		if err != nil {
-			return err
-		}
-		err = pbkt.DeleteBucket(workBkt)
-		if err != nil {
-			return err
-		}
-		err = pbkt.DeleteBucket(jobBkt)
-		if err != nil {
-			return err
-		}
-		err = pbkt.DeleteBucket(paymentBkt)
-		if err != nil {
-			return err
-		}
-		err = pbkt.DeleteBucket(paymentArchiveBkt)
-		if err != nil {
-			return err
-		}
-		err = pbkt.Delete(lastPaymentHeight)
-		if err != nil {
-			return err
-		}
-		err = pbkt.Delete(lastPaymentPaidOn)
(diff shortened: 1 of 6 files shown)


jholdstock:
Pushed an extra commit which adds some tests for the new DB accessors. This is now ready to review.

pool/boltdb.go
```diff
@@ -319,7 +228,40 @@ func nanoToBigEndianBytes(nano int64) []byte {
 	return b
 }
 
-func (db *BoltDB) persistPoolMode(mode uint32) error {
+func (db *BoltDB) fetchSoloPoolMode() (bool, error) {
+	// SoloPoolMode is stored as a uint32 for historical reasons.
```
dnldd, line 232: Probably need to work on this again at some point. Currently not being particularly efficient with the data type being used to store a bool but it works.

dnldd: Just realized postgres support will allow additional pool modes (for example a pool instance that is part of a cluster) so we can tackle this when that is implemented.

pool/hub.go
```diff
@@ -259,18 +259,33 @@ func NewHub(cancel context.CancelFunc, hcfg *HubConfig) (*Hub, error) {
 	}
 	h.chainState = NewChainState(sCfg)
 
-	mode := uint32(0)
+	// Ensure database is in the correct mode.
+	dbMode, err := hcfg.DB.fetchSoloPoolMode()
+	if err != nil {
+		// If pool mode is not set, assume this is a new database and insert
+		// pool mode now.
```
dnldd, line 266: s / `... assume this is a new database and insert pool mode now` / `...assume the database is new and persist the pool mode.`

dnldd (changes requested):
Minor issues, looks good overall.

pool/boltdb.go
jholdstock: Using a boolean is not a good idea if we expect pool mode to have >2 values in future. I think its better to revert back to a number, but rather than using uint32 we should have named with specific values, rather than a naked `0/1`. Maybe like this?

```
type PoolMode uint32

const (
	SoloPool   PoolMode = 0
	SharedPool PoolMode = 1
)

func setPoolMode(PoolMode)

func fetchPoolMode() (PoolMode, error)
```

dnldd: I was thinking of a `mode` field with options like `solo`, `public`, `cluster`. Its more descriptive than using number values.

pool/boltdb.go
```diff
@@ -243,29 +243,41 @@ func (db *BoltDB) fetchSoloPoolMode() (bool, error) {
 		if b == nil {
 			return dbError(ErrValueNotFound, "no pool mode found")
 		}
-		mode = binary.LittleEndian.Uint32(b)
+		dbMode = binary.LittleEndian.Uint32(b)
 		return nil
 	})
 	if err != nil {
-		return false, err
+		return "", err
```
dnldd, line 250: Need to have an unknown state as well instead of falling back on an empty string I think

dnldd (approved):

PR #270, opened 2020-10-25
multi: update harness startup.

dnldd:
**depends on #271**

This adds `--gencertsonly` option as a config option to allow for the generation of needed keypairs without starting the pool. This option is now being used by the harness to generate client ca files before the pool wallet starts up.


jholdstock (commented):
If generating the certs is now a necessary step to deploy dcrpool, should we include something about this in README.md?

dnldd:
Good catch, will update that soon.

jholdstock:
You seem to have your #271 changes here as well

README.md
```diff
@@ -147,6 +147,19 @@ With multiple pool fee addresses set, the mining pool picks one at random for
...
 
+## Wallet Client Authentication
+
+dcrwallet v1.6 requires client authentication certificates to be  provided 
+on startup via the client CA file config option (`--clientcafile`). 
+Since dcrpool is expected to maintain a grpc connection to the wallet it needs 
+to generate the needed certificate before the wallet is started. A config 
+option (`--gencertsonly`) which allows the generation of all key pairs without 
+starting the pool has been added for this purpose. Pool operators running a 
+publicly avilable mining pool will be required to first run their pools
```
jholdstock, line 158:
```suggestion
publicly available mining pool will be required to first run their pools 
```

README.md
```diff
@@ -147,6 +147,19 @@ With multiple pool fee addresses set, the mining pool picks one at random for
 every payout made. The addresses generated from the fee account should be 
 set as the pool fee addresses (`--poolfeeaddrs`) of the mining pool.
 
+## Wallet Client Authentication
+
+dcrwallet v1.6 requires client authentication certificates to be  provided
```
jholdstock, line 152:
```suggestion
dcrwallet v1.6 requires client authentication certificates to be provided 
```

README.md
```diff
@@ -147,6 +147,19 @@ With multiple pool fee addresses set, the mining pool picks one at random for
...
+dcrwallet v1.6 requires client authentication certificates to be  provided 
+on startup via the client CA file config option (`--clientcafile`). 
+Since dcrpool is expected to maintain a grpc connection to the wallet it needs 
+to generate the needed certificate before the wallet is started. A config 
+option (`--gencertsonly`) which allows the generation of all key pairs without 
+starting the pool has been added for this purpose. Pool operators running a 
+publicly avilable mining pool will be required to first run their pools 
+with `--gencertsonly` to generate required key pairs before configuring their 
+pool wallets and starting them. The test harness, [harness.sh](./harness.sh), 
+provides a detailed example for referrence.
```
jholdstock, line 161:
```suggestion
provides a detailed example for reference. 
```

harness.sh
```diff
@@ -238,7 +245,7 @@ y
 ${MASTER_WALLET_SEED}
 EOF" C-m
 sleep 1
-tmux send-keys "dcrwallet -C mwallet.conf " C-m
+tmux send-keys "dcrwallet -C mwallet.conf --debuglevel=info" C-m
```
jholdstock, line 248: `info` is already the default log level, we dont need to set the flag here.

jholdstock (changes requested):

harness.sh
dnldd: explicitly stating that there like the rest of the components, makes it easy to switch debug levels as well by editing that.

jholdstock (approved):

jholdstock:
Merging this so I can rebase and test #269

PR #272, opened 2020-10-26
Improve test coverage for DB accessors

jholdstock:
- Move int64 <> byte conversion to boltdb.go because it is a generic bbolt helper used in multiple places.
- Various improvements to database accessor test coverage.  eg. checking errors more strictly, checking data fields are stored and retrieved correctly.
- Add fetchShare to Database interface. This is only used by test code, but its going to be needed for both BoltDB and PostgresDB, so it should be included in the interface.
- When inserting data, ensure database keys have not already been used. Return ErrValueFound if they have. Postgres will do this at the DB level, so enforcing the same checks in our BoltDB code helps to keep the interface and implementations consistent.

Improving the DB test coverage will help to ensure a Postgres DB implementation matches the Bolt DB implementation (#257)

### DIFF
--- a/pool/account.go
+++ b/pool/account.go
@@ -38,7 +38,8 @@ func NewAccount(address string) *Account {
 	}
 }
 
-// fetchAccount fetches the account referenced by the provided id.
+// fetchAccount fetches the account referenced by the provided id. Returns
+// an error if the account is not found.
 func (db *BoltDB) fetchAccount(id string) (*Account, error) {
 	const funcName = "fetchAccount"
 	var account Account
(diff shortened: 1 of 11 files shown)


jholdstock:
@dnldd  I have deliberately opened a PR with a failing test because I think its something you should see.

I have modified share.go so that it returns an error if you attempt to store two shares with the same ID. This has caused a failing test because the test is trying to insert 5 shares with the same ID.

The previous implementation before I added the error was silently overwriting the value in the database 5 times, which means it was basically having zero effect on anything. What do you think we should do to resolve this? I guess either write 5 shares with different IDs, or just write the share once?

dnldd:
Yeah, that doesn't look right. The share created on times were supposed to be incremented in the loop to generate different share ids. updated the share generation snippet at the reported error line to:


```go
// Ensure payment maturity works as expected.
for i := 0; i < shareCount; i++ {
	// Create readily available shares for account X.
	err = persistShare(db, xID, weight, thirtyBefore+int64(i))
	if err != nil {
		t.Fatal(err)
	}
}
sixtyAfter := time.Now().Add((time.Second * 60)).UnixNano()
for i := 0; i < shareCount; i++ {
	// Create future shares for account Y.
	err = persistShare(db, yID, weight, sixtyAfter+int64(i))
	if err != nil {
		t.Fatal(err)
	}
}
```

That should do it.

jholdstock:
Added your diff and the test is now passing - thanks.

This is now good for review.

dnldd (approved):
Looks good.

PR #275, opened 2020-10-28
pool: upgrade share struct.

dnldd:
This upgrades the share struct to track its created on field directly. Database upgrade and test upgrade implementations have been added accordingly.


davecgh (approved):

jholdstock (approved):
Code looks good and it works for me. Thanks for getting this done so quickly @dnldd

PR #276, opened 2020-10-28
pool: upgrade payment struct. 

dnldd:
This upgrades the payment struct to track its uuid directly. Database upgrade and test upgrade implementations have been added accordingly.


jholdstock (approved):

jholdstock:
Great stuff, thanks!

PR #277, opened 2020-10-29
Payment ID followup

jholdstock:
Following up from #276, a couple of additional changes.

- The first commit in this PR fixes a bug where archived payments were not having their UUID set correctly, the old value was being used. Using `NewPayment` to create the archive payment ensures all fields are set correctly, and also removes the need to call `time.Now` in the database code (duplicated in NewPayment).

- The second commit removes a lot of instances where we were recreating a payment ID with `paymentID()`, even though we already had it accessible in `payment.UUID`


dnldd (approved):
Looks good.

PR #279, opened 2020-10-30
Remove database cleanup from payment manager tests.

jholdstock:
First commit is fixing an existing bug in `paymentmgr_test.go`.

Second is breaking down the monolithic paymentmgr test in to smaller chunks. This enables more detail about test failures, and also removes database cleanup code from within the tests which will help with testing against different database implementations. (#257)

A third commit to only insert accounts into the test database when required. Previously 2 test accounts were inserted every time the test db was initialized, even though the accounts are only required for a small set of tests.


dnldd (approved):
Nicely done, looks good.

PR #280, opened 2020-10-30
Bundled test improvements

jholdstock:
This PR contains multiple small improvements to testing code, most importantly improved test coverage for database code (supporting #257)

- Extra test coverage for DB code in `payments.go` and `acceptedwork.go`
- Properly check type of some errors in tests
- Remove helpers `persistPayment` and `persistAcceptedWork` which didnt really help much.
- Update Database interface so `deleteAcceptedWork` accepts an `id string` param, rather than an `AcceptedWork` struct.

### DIFF
--- a/pool/chainstate_test.go
+++ b/pool/chainstate_test.go
@@ -118,8 +118,8 @@ func testChainState(t *testing.T) {
 		t.Fatalf("expected a valid accepted work, got: %v", err)
 	}
 	_, err = db.fetchAcceptedWork(workB.UUID)
-	if err == nil {
-		t.Fatal("expected a no value found error")
+	if !errors.Is(err, ErrValueNotFound) {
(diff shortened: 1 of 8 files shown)


pool/chainstate_test.go
```diff
@@ -3,6 +3,7 @@ package pool
 import (
 	"context"
 	"encoding/hex"
+	"errors"
```
dnldd, line 6: wondering what triggered this addition considering the change made does not use anything from the `errors` package.

jholdstock: Using `errors.Is` to confirm correct error types. Maybe the import was included in the wrong commit because I commited things in the wrong order.

dnldd: yeah must have slipped through when you were splitting the work done into task focused commits.

dnldd (approved):

PR #281, opened 2020-10-31
Extract DB backup from hub.go

jholdstock:
- Extract DB backup from hub.go. This is a bbolt specific action, and hub should not be aware of the specific database implementation. This will allow hub.go tests to be run against both postgres and bolt implementations. (#257)
-  Hoist db init and shutdown up to main func. With this change, all pool components are now handling a Database interface and have no knowledge of the specific implementation.
- Also adding some non UTF-8 bytes to sample CSRF secret in tests. This is something which caught me out during Postgres implementation.  BoltDB is happy to accept invalid UTF-8 bytes, postgres is not.


dcrpool.go
```diff
@@ -250,9 +250,15 @@ func main() {
 		}
 	}()
 
-	p, err := newPool(cfg)
+	db, err := pool.InitBoltDB(cfg.DBFile)
+	if err != nil {
+		mpLog.Errorf("Failed to initialize database: %v", err)
```
dnldd, line 255: let's not capitalize the errors, to be consistent with the rest of the error definitions.

dcrpool.go
```diff
@@ -250,9 +250,15 @@ func main() {
...
+		os.Exit(1)
+	}
+
+	p, err := newPool(db, cfg)
 	if err != nil {
-		mpLog.Error(err)
+		mpLog.Errorf("Failed to initialize pool: %v", err)
```
dnldd, line 261: Same here.

dcrpool.go
```diff
@@ -289,4 +295,13 @@ func main() {
...
 	p.gui.Run(p.ctx)
 	p.hub.Run(p.ctx)
+
+	// hub.Run() blocks until the pool is fully shut down. When it returns,
+	// close the DB and write a backup file.
+	db.Close()
+	mpLog.Tracef("Backing up database.")
+	err = db.Backup(dbBackupFile)
+	if err != nil {
+		mpLog.Errorf("Failed to write database backup file: %v", err)
```
dnldd, line 305: same here.

dcrpool.go
```diff
@@ -26,6 +26,11 @@ import (
 	"google.golang.org/grpc/credentials"
 )
 
+const (
+	// dbBackupFile is the database backup file name.
+	dbBackupFile = "backup.kv"
```
dnldd, line 31: Since backups are reserved for just th bbolt db implementation should this not be defined there?

jholdstock: Fair point yes. My instinct was to hoist it up the hard-coded string so it could eventually be moved into the config file, but in this case I guess there isnt much point in that. Restored

dcrpool.go
jholdstock: Good catch. Fixed these and a couple of extra ones

jholdstock:
I had to push one further commit which closes the database after writing the backup file, and not before. This is because writing the backup requires the database to still be open.

This maintains the existing behaviour from before this PR.

dnldd (approved):
Looks good.

PR #284, opened 2020-11-05
Introduce an errors package

jholdstock:
This change supports extracting database code into a database package. Both the pool package and database package would be dependent on errors.go, and so a circular dependency would be introduced unless an errors package is introduced.


pool/boltupgrades.go
```diff
@@ -36,10 +36,11 @@ const (
...
 
-	// DBVersion is the latest version of the database that is understood by the
-	// program. Databases with recorded versions higher than this will fail to
-	// open (meaning any upgrades prevent reverting to older software).
-	DBVersion = paymentUUIDVersion
+	// LatestBoltDBVersion is the latest version of the bolt database that is
+	// understood by the program. Databases with recorded versions higher than
+	// this will fail to open (meaning any upgrades prevent reverting to older
+	// software).
+	LatestBoltDBVersion = paymentUUIDVersion
```
dnldd, line 44: This should probably just be `BoltDBVersion`.

errors/error.go
```diff
@@ -0,0 +1,194 @@
...
+func Is(err, target error) bool {
+	return errors.Is(err, target)
+}
+
+// As attempts to assign the error pointed to by target with the first error in
+// err's error chain with a compatible type.  Returns true if target is
+// assigned.
+func As(err error, target interface{}) bool {
+	return errors.As(err, target)
+}
```
dnldd, line 175: Hmm, looking at the call sites for these i think a namespace alias is a better fit than these funcs.

pool/paymentmgr.go
```diff
@@ -29,7 +30,7 @@ const (
 
 	// maxRoundingDiff is the maximum amount of atoms the total
 	// output value of a transaction is allowed to be short of the
-	// provided input due to rounding errors.
+	// provided input due to rounding e.
```
dnldd, line 33: this should be reverted.

pool/paymentmgr.go
```diff
@@ -607,20 +608,20 @@ func (pm *PaymentMgr) generatePayoutTxDetails(ctx context.Context, txC TxCreator
 
 	// Ensure the transaction outputs do not source more value than possible
 	// from the provided inputs and also are consuming all of the input
-	// value after rounding errors.
+	// value after rounding e.
```
dnldd, line 611: same here.

pool/paymentmgr_test.go
```diff
@@ -771,14 +772,14 @@ func testPaymentMgrPayment(t *testing.T) {
...
-	if !errors.Is(err, errors.CreateTx) {
+	if !errors.Is(err, errs.CreateTx) {
 		cancel()
 		t.Fatalf("expected an input output mismatch error")
 	}
 
 	// Ensure generating payout tx details returns an error if the outputs of
 	// the transaction do not exhaust all remaining input value after rounding
-	// errors.
+	// e.
```
dnldd, line 782: same here.

dnldd (approved):
Looks good.

PR #285, opened 2020-11-06
Introduce mock database for testing components

jholdstock:
This PR introduces a mock database generated with https://github.com/golang/mock, and uses it in `testPaymentMgrPPS` and `testPaymentMgrPPLNS`.

As a result, these tests no longer depend on a real database implementation and so can be run as standalone tests rather than as subtests of `pool_test.go`.

Opening as a draft to gather feedback before spending (and possibly wasting) more effort on this. @dnldd do you think this is a direction we should continue in?


dnldd:
As discussed I think we are better off without a dep for this. Better off mocking the db interface for the payment manager tests.

jholdstock:
Closing this in favour of writing our own mocks.

PR #287, opened 2020-11-13
multi: fix lint issues.

dnldd:
This fixes lint outstanding lint issues and wraps errors created by the postgres db driver in order to leverage errors.Is/As functionality.


dajohi (approved):

errors/error.go
```diff
@@ -27,12 +27,27 @@ const (
...
+	DBClose = ErrorKind("ErrDBClose")
+
 	// DBUpgrade indicates a database upgrade error.
 	DBUpgrade = ErrorKind("ErrDBUpgrade")
 
 	// PersistEntry indicates a database persistence error.
 	PersistEntry = ErrorKind("ErrPersistEntry")
 
+	// CreateTable indicates a database table creation error.
+	CreateTable = ErrorKind("CreateTable")
```
jholdstock, line 40: CreateTable is specific to SQL and does not apply to Bolt. As we are aiming to make our DB interface agnostic of the implementation, the interface should not include implementation specific errors. 

This also applies to some existing errors - eg. `ErrBucketCreate`  does not apply to SQL, only Bolt.

We should combine these into something more generic, perhaps something like `ErrCreateStorage`

errors/error.go
```diff
@@ -27,12 +27,27 @@ const (
...
+
+	// BeginTx indicates a database transaction begin error.
+	BeginTx = ErrorKind("BeginTx")
```
jholdstock, line 43: All of these tx errors can probably just be `ErrPersistEntry`. Right now, the calling doesn't need to know the precise nature of the failure, its enough just to know that something went wrong. The nested error generated by the SQL library will contain more details for the log file.

jholdstock (changes requested):

errors/error.go
```diff
@@ -16,119 +16,122 @@ const (
 	// ------------------------------------------
 
 	// ValueNotFound indicates no value found.
-	ValueNotFound = ErrorKind("ErrValueNotFound")
+	ValueNotFound = ErrorKind("ValueNotFound")
 
 	// BucketNotFound indicates no bucket found.
-	BucketNotFound = ErrorKind("ErrBucketNotFound")
+	BucketNotFound = ErrorKind("BucketNotFound")
```
jholdstock, line 22: Maybe rename to `StorageNotFound`?

pool/postgres.go
```diff
@@ -44,53 +44,66 @@ func InitPostgresDB(host string, port uint32, user, pass, dbName string) (*Postg
 		return nil, errs.DBError(errs.DBOpen, desc)
 	}
 
+	makeErr := func(funcName string, table string, err error) error {
```
jholdstock, line 47: passing `funcName` as a param is unnecessary - it is already in scope for the body of this func.

pool/postgres.go
```diff
@@ -219,7 +253,9 @@ func (db *PostgresDB) fetchCSRFSecret() ([]byte, error) {
 			return nil, errs.DBError(errs.ValueNotFound, desc)
 		}
 
-		return nil, err
+		desc := fmt.Sprintf("%s: unable to fetch CSRF secret: %v",
+			funcName, err)
+		return nil, errs.DBError(errs.PersistEntry, desc)
```
jholdstock, line 258: We're fetching here, not persisting

pool/postgres.go
```diff
@@ -234,28 +270,51 @@ func (db *PostgresDB) fetchCSRFSecret() ([]byte, error) {
...
 	tx, err := db.DB.Begin()
 	if err != nil {
 		return err
 	}
 
 	_, err = tx.Exec(insertLastPaymentHeight, height)
 	if err != nil {
-		tx.Rollback()
-		return err
+		rErr := tx.Rollback()
```
jholdstock, line 294: Nice catch finding this, I didnt notice that `Rollback` could return an error.

I'm not sure about the way it is handled though. If we have a `tx.Exec` error and a `tx.Rollback` error, only the `Rollback` error is being returned and logged. The `Exec` error, which could be the root cause of the issue, will be lost. I think we need to log both.

pool/postgres.go
```diff
@@ -234,28 +270,51 @@ func (db *PostgresDB) fetchCSRFSecret() ([]byte, error) {
...
+		if rErr != nil {
+			desc := fmt.Sprintf("%s: unable to rollback persisting last "+
+				"payment height tx: %v", funcName, rErr)
+			return errs.DBError(errs.PersistEntry, desc)
+		}
+
+		desc := fmt.Sprintf("%s: unable to persist last payment height: %s",
+			funcName, err.Error())
```
jholdstock, line 302: Why use `"%s", err.Error()` here and in the following instance, when `"%v", err` is used everywhere else?

pool/postgres.go
```diff
@@ -353,15 +424,22 @@ func (db *PostgresDB) fetchAccount(id string) (*Account, error) {
...
 	return &Account{uuid, address, createdOn}, nil
 }
 
 // deleteAccount purges the referenced account from the database.
 func (db *PostgresDB) deleteAccount(id string) error {
+	const funcName = "deleteAccount"
 	_, err := db.DB.Exec(deleteAccount, id)
-	return err
+	if err != nil {
+		desc := fmt.Sprintf("%s: unable to delete account: %v", funcName, err)
```
jholdstock, line 439: include account id?

pool/postgres.go
```diff
@@ -353,15 +424,22 @@ func (db *PostgresDB) fetchAccount(id string) (*Account, error) {
...
+		return errs.DBError(errs.FetchEntry, desc)
```
jholdstock, line 440: Deleting here, not fetching

pool/postgres.go
```diff
@@ -406,24 +484,35 @@ func (db *PostgresDB) PersistPayment(p *Payment) error {
...
 }
 
 // deletePayment purges the referenced payment from the database. Note that
 // archived payments cannot be deleted.
 func (db *PostgresDB) deletePayment(id string) error {
+	const funcName = "deletePayment"
 	_, err := db.DB.Exec(deletePayment, id)
-	return err
+	if err != nil {
+		desc := fmt.Sprintf("%s: unable to delete payment: %v", funcName, err)
```
jholdstock, line 512: include id?

pool/postgres.go
```diff
@@ -432,13 +521,22 @@ func (db *PostgresDB) ArchivePayment(p *Payment) error {
...
+		return errs.DBError(errs.PersistEntry, desc)
 	}
 
 	_, err = tx.Exec(deletePayment, p.UUID)
 	if err != nil {
-		tx.Rollback()
-		return err
+		rErr := tx.Rollback()
+		if rErr != nil {
+			desc := fmt.Sprintf("%s: unable to rollback delete payment tx: %v",
```
jholdstock, line 533: As above, we may lose the root cause error here

pool/postgres.go
```diff
@@ -618,7 +758,9 @@ func (db *PostgresDB) fetchAcceptedWork(id string) (*AcceptedWork, error) {
 			return nil, errs.DBError(errs.ValueNotFound, desc)
 		}
 
-		return nil, err
+		desc := fmt.Sprintf("%s: unable to fetch accepted work: %v",
```
jholdstock, line 761: id?

pool/postgres.go
```diff
@@ -674,18 +820,26 @@ func (db *PostgresDB) updateAcceptedWork(work *AcceptedWork) error {
 
 // deleteAcceptedWork removes the associated accepted work from the database.
 func (db *PostgresDB) deleteAcceptedWork(id string) error {
+	const funcName = "deleteAcceptedWork"
 	_, err := db.DB.Exec(deleteAcceptedWork, id)
-	return err
+	if err != nil {
+		desc := fmt.Sprintf("%s: unable to delete account work: %v",
```
jholdstock, line 826: id?

pool/postgres.go
```diff
@@ -737,20 +895,33 @@ func (db *PostgresDB) persistJob(job *Job) error {
...
 	return nil
 }
 
 // deleteJob removes the associated job from the database.
 func (db *PostgresDB) deleteJob(id string) error {
+	const funcName = "deleteJob"
 	_, err := db.DB.Exec(deleteJob, id)
-	return err
+	if err != nil {
+		desc := fmt.Sprintf("%s: unable to delete job entry: %v",
```
jholdstock, line 909: id?

jholdstock (changes requested):

pool/postgres.go
```diff
@@ -234,28 +270,53 @@ func (db *PostgresDB) fetchCSRFSecret() ([]byte, error) {
...
+		rErr := tx.Rollback()
+		if rErr != nil {
+			desc := fmt.Sprintf("%s: unable to rollback persist last payment "+
+				"paid on time tx: %v, initial error: %v", funcName,
+				rErr, err)
+			return errs.DBError(errs.PersistEntry, desc)
+		}
+
+		desc := fmt.Sprintf("%s: unable to persist last payment paid on "+
+			"time: %s", funcName, err.Error())
```
jholdstock, line 318: `"%v", err`

pool/postgres.go
```diff
@@ -449,29 +552,48 @@ func (db *PostgresDB) ArchivePayment(p *Payment) error {
 		aPmt.CreatedOn, aPmt.PaidOnHeight, aPmt.TransactionID, aPmt.Source.BlockHash,
 		p.Source.Coinbase)
 	if err != nil {
-		tx.Rollback()
-		return err
+		rErr := tx.Rollback()
+		if rErr != nil {
+			desc := fmt.Sprintf("%s: unable to rollback archived payment "+
```
jholdstock, line 557: Missing initial error

jholdstock (changes requested):
Two more things and this is good to go.

jholdstock (approved):

PR #288, opened 2020-11-13
test: add sql linters

dajohi:


run_tests.sh
```diff
@@ -47,4 +47,6 @@ golangci-lint run --disable-all --deadline=10m \
   --enable=vet \
   --enable=gosimple \
   --enable=unconvert \
-  --enable=ineffassign
+  --enable=ineffassign \
+  --enable=rowserrcheck \
```
jholdstock, line 51: I just tested this and `rowserrcheck` doesn't actually work on our code. Probably because our calls to `db.Query` are in different funcs to our calls to `rows.Next()` and `rows.Err()`.

No harm adding it anyway. Might be useful someday in the future.

run_tests.sh
```diff
@@ -47,4 +47,6 @@ golangci-lint run --disable-all --deadline=10m \
...
+  --enable=sqlclosecheck
```
jholdstock, line 52: Lol... `sqlclosecheck` doesnt seem to work either. We don't explicity close rows anywhere in our code, yet this linter does not complain.

I even prevented automatic closure by only reading the first row in a set and leaving the rest unread, the linter still did not complain.

Again - I guess there is no harm in adding it anyway?

dnldd: Yeah I tried this, no dice. we are not explicitly closing rows currently so there should be issues reported. @dajohi how do we get this to work?

PR #289, opened 2020-11-23
multi: update harness db startup.

dnldd:
This updates the pool startup process to purge an existing postgres db on startup for simnet testing purposes.

### DIFF
--- a/harness.sh
+++ b/harness.sh
@@ -31,7 +31,7 @@ LAST_N_PERIOD=5m
 GUI_DIR="${HARNESS_ROOT}/gui"
 
 # Using postgres requires the DB specified below to exist and contain no data.
-USE_POSTGRES=true
+USE_POSTGRES=1
 POSTGRES_HOST=127.0.0.1
 POSTGRES_PORT=5432
 POSTGRES_USER=dcrpooluser
@@ -74,7 +74,7 @@ CLIENT_ADDRS=(
 
 # Number of mining clients to create. Maximum is determined by number of client
 # addresses above - currently 20.
-NUMBER_OF_CLIENTS=2
+NUMBER_OF_CLIENTS=1
 
 if [ -d "${HARNESS_ROOT}" ]; then
   rm -R "${HARNESS_ROOT}"
@@ -143,6 +143,7 @@ postgresport=${POSTGRES_PORT}
 postgresuser=${POSTGRES_USER}
 postgrespass=${POSTGRES_PASS}
 postgresdbname=${POSTGRES_DBNAME}
+purgedb=1
 EOF
 
 cat > "${HARNESS_ROOT}/mwallet/dcrmwctl.conf" <<EOF
@@ -205,7 +206,7 @@ echo "Starting simnet master node"
 tmux send-keys "dcrd --appdata=${HARNESS_ROOT}/master \
 --rpcuser=${RPC_USER} --rpcpass=${RPC_PASS} \
 --miningaddr=${POOL_MINING_ADDR} \
---txindex --debuglevel=info \
+--txindex --debuglevel=debug \
 --simnet" C-m
 
 ################################################################################
--- a/harness.sh
+++ b/harness.sh
@@ -30,7 +30,7 @@ PAYMENT_METHOD="pplns"
 LAST_N_PERIOD=5m
 GUI_DIR="${HARNESS_ROOT}/gui"
 
-# Using postgres requires the DB specified below to exist and contain no data.
+# Using postgres requires the DB specified below to already exist.
 USE_POSTGRES=1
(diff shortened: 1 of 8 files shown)


pool/postgres.go
```diff
@@ -100,6 +109,18 @@ func (db *PostgresDB) Close() error {
 	return nil
 }
 
+// Purge wipes all persisted data.
+func (db *PostgresDB) purge() error {
```
jholdstock, line 113: On the line above and below, `purge` has a capital `P`

pool/postgres.go
```diff
@@ -35,6 +35,15 @@ func InitPostgresDB(host string, port uint32, user, pass, dbName string) (*Postg
 		return nil, errs.DBError(errs.DBOpen, desc)
 	}
 
+	pdb := &PostgresDB{db}
+	if purgeDB {
+		err := pdb.purge()
+		if err != nil {
+			desc := fmt.Sprintf("%s: unable to purge db: %v", funcName, err)
```
jholdstock, line 42: `purge()` is already wrapping the error with `errs.DeleteEntry`. Don't need to wrap it again here.

pool/postgres.go
```diff
@@ -100,6 +109,18 @@ func (db *PostgresDB) Close() error {
 	return nil
 }
 
+// Purge wipes all persisted data.
```
jholdstock, line 112: Mention here that this should be used on simnet only.

pool/sql_queries.go
```diff
@@ -73,6 +73,15 @@ const (
 		confirmed BOOLEAN NOT NULL
 	);`
 
+	purgeDB = `DROP TABLE IF EXISTS
```
jholdstock, line 76: This query already exists in `pool_test.go` and shouldn't be duplicated. We probably need to remove the other one and keep this one.

harness.sh
```diff
@@ -143,6 +143,7 @@ postgresport=${POSTGRES_PORT}
 postgresuser=${POSTGRES_USER}
 postgrespass=${POSTGRES_PASS}
 postgresdbname=${POSTGRES_DBNAME}
+purgedb=1
```
jholdstock, line 146: I dont think `purgedb` should be set permanently in the config file. My workflow often looks like this:

1. Start the harness
1. Test dcrpool features
1. Make code changes
1. Rebuild dcrpool
1. Restart dcrpool in tmux
1. Goto 2.

This workflow wont be possible if dcrpool purges every time it starts up.

Perhaps we should set `--purgedb` on the CLI instead, towards the bottom of this file, so it is only used the first time?

harness.sh
```diff
@@ -74,7 +74,7 @@ CLIENT_ADDRS=(
 
 # Number of mining clients to create. Maximum is determined by number of client
 # addresses above - currently 20.
-NUMBER_OF_CLIENTS=2
+NUMBER_OF_CLIENTS=1
```
jholdstock, line 77: Deliberate change?

jholdstock (changes requested):
A few little changes suggested, and also a bigger idea - if we add `purge` to the database interface, we would no longer need `teardownPostgresDB` and `teardownBoltDB` in the `pool_test.db`. We could instead just call `db.purge`.

harness.sh
dnldd: Yeah the harness only needs a miner to work so it should probably start from there till more are needed for specific test cases.

pool/boltdb.go
```diff
@@ -424,6 +424,106 @@ func (db *BoltDB) Close() error {
 	return db.DB.Close()
 }
 
+// purge wipes all persisted data. This is intended for use with tesnet and
```
jholdstock, line 427: testnet*

jholdstock, line 427: testnet*

pool/postgres.go
```diff
@@ -109,9 +108,10 @@ func (db *PostgresDB) Close() error {
 	return nil
 }
 
-// Purge wipes all persisted data.
+// purge wipes all persisted data. This is intended for use with tesnet and
```
jholdstock, line 111: testnet*

harness.sh
```diff
@@ -31,7 +31,7 @@ LAST_N_PERIOD=5m
 GUI_DIR="${HARNESS_ROOT}/gui"
 
 # Using postgres requires the DB specified below to exist and contain no data.
```
jholdstock, line 33: This comment is no longer true. Previously the database had to be empty, but now it does not because it will be automatically purged.

jholdstock (changes requested):
Code looking good now, just some minor comment issues to fix

jholdstock (approved):

PR #296, opened 2020-12-22
Marshal records before opening DB transaction

jholdstock:


jholdstock:
After chatting with @dnldd and discovering that `json.Marshal` has practically zero chance of returning an error, there is nothing of value in this PR.

PR #297, opened 2020-12-23
multi: various improvements

dnldd:
This improves config options handling, error outputs and letsencrypt http redirects.


gui/gui.go
```diff
@@ -266,27 +266,26 @@ func (ui *GUI) Run(ctx context.Context) {
...
 				Prompt:     autocert.AcceptTOS,
-				Cache:      certCache,
+				Cache:      autocert.DirCache("certs"),
 				HostPolicy: autocert.HostWhitelist(ui.cfg.Domain),
 			}
 
-			// Ensure port 80 is not already in use.
-			port80 := ":80"
-			listener, err := net.Listen("tcp", port80)
+			// Ensure the gui port is not already in use.
```
jholdstock, line 275: When we discussed this on Matrix, I thought the idea was to remove the http redirect completely, rather than just changing it from port 80 to the port provided in config. Can we remove it completely? I don't think its really needed, all clients should be expected to use https.

config.go
```diff
@@ -557,33 +560,61 @@ func loadConfig() (*config, []string, error) {
...
+		fmt.Fprintln(os.Stderr, usageMessage)
 		return nil, nil, err
 	}
 
 	// Create the data directory.
 	err = os.MkdirAll(cfg.DataDir, 0700)
 	if err != nil {
-		str := "%s: failed to create data directory: %v"
-		err := fmt.Errorf(str, funcName, err)
+		str := "%s: unable to create data directory (%s): %v"
```
jholdstock, line 588: `funcName` is used in a handful of places, but not all. Any reason? I guess we could remove it completely

cmd/miner/main.go
```diff
@@ -22,7 +21,6 @@ func main() {
 	// and configures it accordingly.
 	cfg, _, err := loadConfig()
 	if err != nil {
-		fmt.Println(err)
 		return
```
jholdstock, line 24: os.Exit(1) here would be better than returning and exiting with code 0

jholdstock (changes requested):
Apologies for the delay on this, xmas etc.

A couple of minor things

config.go
dnldd: I think it makes sense to keep it for config related errors resulting from creating directories, generating certificates etc which are not directly related to the config process.Errors related to config parsing and parameter constraints can do without the func name label since those indicate an app-wide config issue that's why `funcName` is used only in some places.

gui/gui.go
dnldd: I agree after taking another look, clients should be expecting https and if the redirect is needed it can be handled by the pool operator's web server. The redirect shouldn't be default as a result.

jholdstock:
All good after fixing that typo

dnldd:
Had to push another commit.

pool/hub.go
```diff
@@ -716,7 +716,7 @@ func (h *Hub) FetchWorkQuotas() ([]*Quota, error) {
 func (h *Hub) AccountExists(accountID string) bool {
 	_, err := h.cfg.DB.fetchAccount(accountID)
 	if err != nil {
-		log.Tracef("Unable to fetch account for id: %s", accountID)
+		log.Errorf("unable to fetch account for id: %s", accountID)
```
jholdstock, line 719: I dont agree with this being an error. This is used by the UI to check if an account ID provided by a site visitor exists - there is not an error or problem if the user provides an invalid ID - they could be on the wrong pool, or made a typo. Trace or Info level maximum imo.

jholdstock (changes requested):

pool/hub.go
dnldd: Noted, reverting.

jholdstock (approved):

PR #298, opened 2021-01-04
Add version command.

jholdstock:
Add a version config item which will cause the application to print version information and then exit.


config.go
```diff
@@ -88,6 +88,7 @@ var runServiceCommand func(string) error
 
 // config defines the configuration options for the pool.
 type config struct {
+	ShowVersion           bool          `long:"version" description:"Display version information and exit"`
```
dnldd, line 91: Missing a period at the end of the description.

dnldd (approved):

PR #299, opened 2021-01-20
gui: fix pagination errors.

### LINKED ISSUE #292: Remove hardcoded localhost
`index.html` connection instructions have a hard-coded localhost. This should be dynamic to display the actual current hostname.

https://github.com/decred/dcrpool/blob/49f5ce59892f6dfaab440d79b3063e21a99e82a4/gui/assets/templates/index.html#L88

dnldd:
This fixes pagination related errors when the data source returns nil. The supported miner list has also been updated.

Closes #292

### DIFF
--- a/gui/index.go
+++ b/gui/index.go
@@ -38,6 +38,11 @@ func (ui *GUI) renderIndex(w http.ResponseWriter, r *http.Request, modalError st
 	// time, but the GUI doesn't use them yet.
 	lastPaymentHeight, _, _ := ui.cache.getLastPaymentInfo()
 
+	address := `127.0.0.1`
+	if ui.cfg.Domain != "" {
+		address = ui.cfg.Domain
+	}
+
 	data := indexPageData{
 		HeaderData: headerData{
 			CSRF:        csrf.TemplateField(r),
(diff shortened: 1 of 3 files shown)


gui/index.go
```diff
@@ -38,6 +38,13 @@ func (ui *GUI) renderIndex(w http.ResponseWriter, r *http.Request, modalError st
 	// time, but the GUI doesn't use them yet.
 	lastPaymentHeight, _, _ := ui.cache.getLastPaymentInfo()
 
+	var address string
+	if ui.cfg.UseLEHTTPS {
```
jholdstock, line 42: Should use `if ui.cfg.Domain != nil` here.

If somebody is using another webserver like nginx to provide TLS, `ui.cfg.UseLeHTTPS` will be `false` and `127.0.0.1` will be incorrectly shown on the website.

jholdstock (changes requested):

gui/index.go
dnldd: Good catch, fixing.

jholdstock: I dont get why you've added the extra logic for UseLEHTTPS? Surely just `if ui.cfg.Domain != ""` is enough?

dnldd: It's for the two cases where letsencrypt is being used and when its not, In the case where its not but TLS is still being provided then the domain would be set. In both cases the domain shown should be the domain set.

dnldd: you're right, no need checking the letsencrypt flag to set the domain.

jholdstock (approved):

PR #304, opened 2021-01-28
multi: stratum improvements.

dnldd:
This extends client read/write timeouts, `mining.subscribe` response handling and other stratum related improvements.


jholdstock:
Deployed to https://pool.jholdstock.uk/

Successfully mining blocks, but still seeing a lot of errors on the pool log:

```
2021-02-10 10:12:34.351 [ERR] POOL: submitted work from d99f78a0/cpu is not less than the network target difficulty
2021-02-10 10:25:12.450 [ERR] POOL: submitted work from d99f78a0/cpu is not less than the network target difficulty
2021-02-10 10:26:33.480 [ERR] POOL: submitted work from d99f78a0/cpu is not less than the network target difficulty
2021-02-10 10:31:27.851 [ERR] POOL: submitted work from d99f78a0/cpu is not less than the network target difficulty
2021-02-10 10:31:34.448 [ERR] POOL: submitted work from d99f78a0/cpu is not less than the network target difficulty
2021-02-10 10:31:41.004 [ERR] POOL: submitted work from d99f78a0/cpu is not less than the network target difficulty
2021-02-10 10:31:47.273 [ERR] POOL: submitted work from d99f78a0/cpu is not less than the network target difficulty
2021-02-10 10:32:11.731 [INF] POOL: Mined work 0000000ce28b245f67ae1103ec8619fce82b41a623d697e4e2689924557e29cb confirmed by connected block #618036
```

pool/client.go
```diff
@@ -762,7 +799,7 @@ func (c *Client) updateWork() {
 		return
 	}
 	workNotif := WorkNotification(job.UUID, prevBlock, genTx1, genTx2,
-		blockVersion, nBits, nTime, true)
+		blockVersion, nBits, nTime, prioritize)
```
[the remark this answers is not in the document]
dnldd: `CleanJob` prioritizes the incoming work notification over the work data being processed by the miner. Better off using the same terminology however, will update.

cmd/miner/client.go
```diff
@@ -131,7 +131,7 @@ func (m *Miner) keepAlive(ctx context.Context) {
 			m.encoder = json.NewEncoder(m.conn)
 			m.reader = bufio.NewReader(m.conn)
 
-			log.Debugf("miner connected to %s", m.config.Pool)
+			log.Tracef("miner connected to %s", m.config.Pool)
```
jholdstock, line 134: I dont think this needs to be relegated to trace, its fine on debug. Tbh I might  even argue that this line would be fine at info level.

Trace should be reserved for things which either log excessive amount of data, or are logged very frequently. This line does neither.

cmd/miner/config.go
```diff
@@ -474,7 +474,7 @@ func loadConfig() (*config, []string, error) {
 	// done. This prevents the warning on help messages and invalid
 	// options. Note this should go directly before the return.
 	if configFileError != nil {
-		log.Debugf("%v", configFileError)
+		log.Tracef("%v", configFileError)
```
jholdstock, line 477: This seems like it should be warn or error level.

pool/client.go
```diff
@@ -864,10 +860,18 @@ func (c *Client) process() {
 
 				case Submit:
 					err := c.handleSubmitWorkRequest(c.ctx, req, allowed)
+					if errors.Is(err, errs.NetworkDifficulty) {
+						// Submissions less than the network difficulty should
+						// not be treated as errors.
+						log.Trace(err)
```
jholdstock, line 866: For the reasons above - I think debug or info are fine here. This is not an excessive amount of info, or logged excessively often

pool/endpoint.go
```diff
@@ -192,7 +194,7 @@ func (e *Endpoint) connect(ctx context.Context) {
 			e.wg.Add(1)
 			go client.run()
 
-			log.Debugf("Mining client connected. extranonce1=%s, addr=%s",
+			log.Tracef("Mining client connected. extranonce1=%s, addr=%s",
```
jholdstock, line 197: Debug or info on this imo.

jholdstock (approved):

PR #305, opened 2021-02-16
pool: fix hash calc spanning multiple cycles.

dnldd:
This fixes a bug related to hash calculations spanning multiple cycles.

### DIFF
--- a/pool/boltdb.go
+++ b/pool/boltdb.go
@@ -1438,9 +1438,9 @@ func (db *BoltDB) fetchHashData(id string) (*HashData, error) {
 }
 
 // listHashData fetches all hash data updated after the provided minimum time.
-func (db *BoltDB) listHashData(minNano int64) (map[string][]*HashData, error) {
+func (db *BoltDB) listHashData(minNano int64) (map[string]*HashData, error) {
 	const funcName = "listHashData"
-	data := make(map[string][]*HashData)
+	data := make(map[string]*HashData)
 
 	err := db.DB.View(func(tx *bolt.Tx) error {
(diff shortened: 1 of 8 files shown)


jholdstock (commented):
Code is good so far, and the pool hashrate calc is working much better than previously, but a couple of things:

- When a client first connects, the pool reports it as 838.86 KH/s for some time before updating to a more realistic figure. Sometimes this lasts a long time.

- When a client timeout occurs (eg. because the client has not submitted anything for a while, example below)  the calculated hashrate for that client is lost, and takes quite some time to reappear. The extent of this problem is such that I have three CPU miners connected to my pool, and for the last 30 minutes I have only seen one or two clients listed at a time on the GUI - I have never seen all three displayed at once. Often there are zero displayed.

        2021-02-16 10:20:20.122 [ERR] POOL: f12dc348/cpu: read timeout: read tcp 127.0.0.1:5550->127.0.0.1:34220: i/o timeout
        2021-02-16 10:20:20.123 [ERR] POOL: f12dc348/cpu: unable to send close message: write tcp 127.0.0.1:5550->127.0.0.1:34220: i/o timeout
        2021-02-16 10:21:20.120 [DBG] POOL: Mining client connected. extranonce1=fe2cc547, addr=127.0.0.1:34398



I think both of these may be solved if miners are submitting more often - are we able to further reduce the pool difficulty when running on testnet?

jholdstock (approved):
Still some issues outstanding but approving this PR on the basis that it is a good improvement over what we currently have, and the issues are likely only with cpu miners and not likely to manifest on real mining hardware.

PR #309, opened 2021-03-12
Set PaidOnHeight & TxID for archived payments.

jholdstock:
Previously these values were not being saved correctly because they were not being set when a new Payment was created in the DB code.


dnldd (approved):
Good catch, tested via the harness.

PR #310, opened 2021-03-12
Disable account lookup on solo pools.

### LINKED ISSUE #308: Disable account lookup in solo pool mode
When running in solo pool mode, the account lookup functionality is still available in the GUI but it does not work. This feature should either be hidden or disabled.

jholdstock:
Disable account lookup form element and display a message to notify the feature is not available.

Closes #308


dnldd (approved):
This will do until the improved navigation is implemented, tested via the harness.

PR #312, opened 2021-03-12
multi: update a homedir config flag.

### LINKED ISSUE #307: --homedir config is not consistent with dcrd
The homedir config is not consistent with dcrd and dcrwallet configs that perform the same duty.

* dcrpool has homedir
* dcrd has appdata
* dcrwallet has appdata

These config names should all be the same across all apps.  This make configuration easier to understand and programmatically configure.

Please deprecate homedir and add appdata config.

dnldd:
This updates the `--homedir` flag to `--appdata` to be consistent with `dcrd` and `dcrwallet`. 

This is a breaking change. Normally deprecation would be preferred but since the upcoming release already has breaking changes 
and it'd not be great managing two homedir config vars with different values.

Resolves #307

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ options will be needed, refer to [postgres.md](./docs/postgres.md).
 ### Example output of a solo pool startup
 
 ```no-highlight
-dcrpool --configfile=pool.conf --homedir=/tmp/dcrpool-harness/pool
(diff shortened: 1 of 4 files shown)


jholdstock (changes requested):
Changes fine so far, but

- README needs updating in two places.
- We might as well update the same thing in `cmd/miner/config.go` as well.

jholdstock (approved):

PR #318, opened 2021-04-07
pool: superfluous work notif after submission.

dnldd:
This removes an unnecessary work notification to clients that submit valid work submission. It's is unnecessary because additional work submissions can be found from the unexplored nonce space and if there is the need for updated work the timestamp rolled work process will provide that.

### DIFF
--- a/pool/client.go
+++ b/pool/client.go
@@ -754,11 +754,18 @@ func (c *Client) updateWork(cleanJob bool) {
 	subscribed := c.subscribed
 	c.statusMtx.RUnlock()
 
+	c.mtx.RLock()
+	id := c.id
+	c.mtx.RUnlock()
+
 	if !subscribed || !authorized {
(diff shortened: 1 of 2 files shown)


pool/client.go
```diff
@@ -754,11 +754,18 @@ func (c *Client) updateWork(cleanJob bool) {
 	subscribed := c.subscribed
 	c.statusMtx.RUnlock()
 
+	c.mtx.RLock()
+	id := c.id
+	c.mtx.RUnlock()
+
 	if !subscribed || !authorized {
+		log.Tracef("can only timestamp-roll work for subscribed " +
```
jholdstock, line 762: I think this, and the other new log line, would be acceptable at debug level? I don't think this will be spammed too often?

jholdstock (approved):

PR #319, opened 2021-04-08
miner: avoid sending already known submissions. 

dnldd:
This addresses some issues identified with the cpu miner in testing. Each issue has been addressed in a commit.

- miner: prioritize clean jobs:  
	- This updates the miner to immediately begin solving new work if it is marked as a clean job.

- miner: avoid sending already known submissions:  
	- This avoids sending already known submissions by comparing the previously sent submission nonce data to the newly solved nonce data and only proceeding if they do not match.

- miner: avoid submitting work when disconnected:  
	- This updates the disconnection process to first update the solver and avoid work submissions.

- miner: fix possible read hang:  
	- This fixes a case where process goroutine terminates on a context signal while the read process sends to its unbuffered read channel.

### DIFF
--- a/cmd/miner/client.go
+++ b/cmd/miner/client.go
@@ -121,6 +121,8 @@ func (m *Miner) keepAlive(ctx context.Context) {
 
 			time.Sleep(time.Second * 2)
 
+			log.Infof("dialing %s...", m.config.Pool)
+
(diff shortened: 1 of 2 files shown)


jholdstock (commented):
I think the new log statements would be acceptable at DEBUG level

jholdstock (approved):

PR #327, opened 2021-05-05
Add SIGTERM and SIGHUP to shutdown signals.

jholdstock:
More info at decred/dcrd#2644

### DIFF
--- a/cmd/miner/main.go
+++ b/cmd/miner/main.go
@@ -12,10 +12,14 @@ import (
 	"runtime"
 )
 
+// signals defines the signals that are handled to do a clean shutdown.
+// Conditional compilation is used to also include SIGTERM and SIGHUP on Unix.
+var signals = []os.Signal{os.Interrupt}
+
 func main() {
 	// Listen for interrupt signals.
 	interrupt := make(chan os.Signal, 1)
-	signal.Notify(interrupt, os.Interrupt)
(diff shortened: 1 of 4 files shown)


cmd/miner/main.go
```diff
@@ -10,12 +10,13 @@ import (
 	"os"
 	"os/signal"
 	"runtime"
+	"syscall"
 )
 
 func main() {
 	// Listen for interrupt signals.
 	interrupt := make(chan os.Signal, 1)
-	signal.Notify(interrupt, os.Interrupt)
```
davecgh, line 18: I'm pretty sure this won't build on windows.  It needs to be a separate file that only builds on unix.

jholdstock: Yep I'm pretty sure you are correct, however, as @chappjc [commented on dcrseeder](https://github.com/decred/dcrseeder/pull/41#discussion_r553845018) some time ago, I don't expect anybody would run dcrpool on Windows.

We can move this into a separate file so it only builds on specific systems, as is done in dcrd and dcrwallet. What do you think @dnldd?

dnldd: Yeah that seems to be the way to go.

jholdstock: Updated

dnldd (approved):

PR #331, opened 2021-05-21
Update to golangci-lint 1.40.1.

jholdstock:
Linter `golint` has been deprecated and replaced with `revive`.


dnldd (approved):
This looks good, will merge after #334 however since that's currently being tested.

PR #333, opened 2021-05-24
Additional trace logging.

jholdstock:
Log lines added whilst debugging some recent issues (miner disconnections not handled smoothly, failedTxConfs exceeding specified maximum)


dnldd:
Do we still need this?

jholdstock:
Probably not.

PR #334, opened 2021-05-26
multi: use getTransaction grpc for confirmCoinbases. 

dnldd:
This reworks coinbase confirmation function to uses the getTransaction grpc to fetch transaction details and confirm them. The rescan begin height is now a coinbase maturity length of blocks from the lowest reported block with a tx having inaccurate confirmation information. Tests have been updated.

### DIFF
--- a/pool/hub.go
+++ b/pool/hub.go
@@ -100,6 +100,7 @@ var (
 type WalletConnection interface {
 	SignTransaction(context.Context, *walletrpc.SignTransactionRequest, ...grpc.CallOption) (*walletrpc.SignTransactionResponse, error)
 	PublishTransaction(context.Context, *walletrpc.PublishTransactionRequest, ...grpc.CallOption) (*walletrpc.PublishTransactionResponse, error)
+	GetTransaction(context.Context, *walletrpc.GetTransactionRequest, ...grpc.CallOption) (*walletrpc.GetTransactionResponse, error)
 	Rescan(ctx context.Context, in *walletrpc.RescanRequest, opts ...grpc.CallOption) (walletrpc.WalletService_RescanClient, error)
 }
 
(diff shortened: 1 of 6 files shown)


pool/paymentmgr.go
```diff
@@ -877,42 +943,66 @@ func (pm *PaymentMgr) payDividends(ctx context.Context, height uint32, treasuryA
...
+		// Do not error if coinbase spendable confirmation requests
+		// are terminated by the context cancellation.
 		if !errors.Is(err, errs.ContextCancelled) {
 			return err
 		}
 
 		return nil
 	}
 
+	// var maxSpendableHeight uint32
```
jholdstock, line 969: If this PR is merged, we obviously need to remove all this commented code first (and any other code which is no longer required).

jholdstock: Also need to remove `v2` from the name `confirmCoinbasesV2`

pool/paymentmgr_test.go
```diff
@@ -639,6 +646,8 @@ func testPaymentMgrPayment(t *testing.T) {
 		cancel()
 	}()
 
+	// TODO (dnldd): add tests for coinfirmCoinbasesV2.
```
jholdstock, line 649: Required before merging

jholdstock: Also ensure tests for v1 are removed.

pool/paymentmgr.go
```diff
@@ -592,6 +596,66 @@ func (pm *PaymentMgr) confirmCoinbases(ctx context.Context, txHashes map[chainha
...
+			}
+
+			// Remove the confirmed transaction.
+			if resp.Confirmations >= maxSpendableConfs {
+				delete(txHashes, keys[idx])
+			}
+		}
+	}
+
+	log.Tracef("%d coinbase(s) are not confirmed:", len(txHashes))
```
jholdstock, line 647: Dont need this logging if `len(txHashes) == 0`

jholdstock (commented):
Added a few comments, mostly for myself to remember I need to re-review some of the points later.

---

The reason we originally used `TxConfNotifications` to track when coinbases become spendable was because of a race condition explained here in #222.

The old `confirmCoinbases` blocked until all of the coinbases were spendable. I don't think the new one is blocked, so it seems like it will be prone to the same race condition.

pool/paymentmgr.go
dnldd: yeah this isn't expected to be merged any time soon, the plan is to test this extensively to confirm it solves the inaccurate confirmation issue at least after a rescan. With that done I can remove the functions no longer needed, update the tests and get it ready for merging.

pool/paymentmgr_test.go
dnldd: noted.

dnldd:
I don't think the issue described in #222 is going to be a problem here since `gettxout` was informing us of what the node knew about the provided tx output not factoring in the wallet's state concerning that output (the wallet could be updated late for example). The current solution being used in `confirmCoinbasesV2` is calling the wallet's `gettransaction` grpc for the transactions of interest. If the wallet doesn't know about them yet it will have two more blocks to do so before a rescan is performed. The disparity we had before concerning the information the wallet and node had at a given time on a tx shouldn't be an issue now since we ask the wallet to confirm txs it will need to create the payout transaction directly, however we'll still have to confirm `gettransaction` works as expected and the expanded rescan range is adequate in mitigating the current tx confirmation issues being had with recieving tx confirmation notifications.

jholdstock:
Understood. Will deploy onto my test site and monitor for a few days.

pool/hub.go
```diff
@@ -244,22 +245,22 @@ func NewHub(cancel context.CancelFunc, hcfg *HubConfig) (*Hub, error) {
...
+		SoloPool:              h.cfg.SoloPool,
+		PaymentMethod:         h.cfg.PaymentMethod,
+		PoolFeeAddrs:          h.cfg.PoolFeeAddrs,
+		WalletAccount:         h.cfg.WalletAccount,
+		WalletPass:            h.cfg.WalletPass,
+		GetBlockConfirmations: h.getBlockConfirmations,
+		FetchTxCreator:        func() TxCreator { return h.nodeConn },
+		FetchTxBroadcaster:    func() TxBroadcaster { return h.walletConn },
+		CoinbaseConfTimeout:   h.cfg.CoinbaseConfTimeout,
+		Cancel:                h.cancel,
```
jholdstock, line 261: Cancel is not used anywhere. Should either be used or removed.

pool/paymentmgr.go
```diff
@@ -488,108 +483,63 @@ func (pm *PaymentMgr) applyTxFees(inputs []chainjson.TransactionInput, outputs m
...
-		if notif.err != nil {
-			desc := fmt.Sprintf("%s: unable to fetch tx confirmation "+
-				"notifications, %s", funcName, notif.err)
-			return nil, errs.PoolError(errs.TxConf, desc)
-		}
-		return notif.resp, nil
-	}
-}
-
 // confirmCoinbases ensures the coinbases referenced by the provided
```
jholdstock, line 486: I think this comment needs updating/expanding. Please explain what confirmCoinbases is doing, how it is doing it, and why it is necessary.

pool/paymentmgr.go
```diff
@@ -877,35 +829,22 @@ func (pm *PaymentMgr) payDividends(ctx context.Context, height uint32, treasuryA
...
-	err = pm.confirmCoinbases(pCtx, inputTxHashes, stopAfter)
+	err = pm.confirmCoinbases(pCtx, txB, inputTxHashes)
 	if err != nil {
-		atomic.AddUint32(&pm.failedTxConfs, 1)
-
-		// Track the transactions with inaccurate confirmation data.
-		pm.mtx.Lock()
-		for k, v := range inputTxHashes {
-			pm.txConfHashes[k] = v
+		if len(inputTxHashes) != 0 {
```
jholdstock, line 834: This `if` seems unnecessary. How can an error occur if there are zero `inputTxHashes`?

jholdstock (changes requested):

pool/hub.go
dnldd: Noted, removing.

pool/paymentmgr.go
dnldd: Sure thing, updated.

pool/paymentmgr.go
dnldd: Yeah, that's no needed. Updating.

jholdstock (approved):

PR #350, opened 2023-09-12
config: Remove unused err return on config parser.

davecgh:
**This is rebased on #349**.

This removes the unused error return from the `newConfigParser` function and updates the callers accordingly.

PR #351, opened 2023-09-12
main: Correct defer handling in main entrypoint.

davecgh:
**This is rebased on #350**.

This refactors the main code into a separate function that returns an error instead of invoking `os.Exit` directly to ensure that the defers that are setup prior to the exits are run as intended.

The `main` func will now exit with the non-zero error code when an error is returned from the refactored separate func.


jholdstock (approved):
Looks good but needs rebase.

PR #352, opened 2023-09-12
config: Consolidate error reporting.

davecgh:
**This requires #351**.

This modifies the initial configuration error handling logic to match dcrd.  In particular, the logic is modified to avoid a bunch of duplicate code and fix some cases that were not logging the errors as expected.

It accomplishes this by consolidating the error reporting in a single place in the caller and introducing a custom error type which signals that the usage message should be suppressed in order to retain the primary behavior the originally was the reason for the prints being in the config function to begin with.

Also, the application name is passed into the function now since it is needed in both places.

### DIFF
--- a/config.go
+++ b/config.go
@@ -171,7 +171,7 @@ func parseAndSetDebugLevels(debugLevel string) error {
 	if !strings.Contains(debugLevel, ",") && !strings.Contains(debugLevel, "=") {
 		// Validate debug log level.
 		if !validLogLevel(debugLevel) {
-			str := "the specified debug level [%v] is invalid"
+			const str = "the specified debug level [%v] is invalid"
 			return fmt.Errorf(str, debugLevel)
 		}
 
@@ -185,7 +185,7 @@ func parseAndSetDebugLevels(debugLevel string) error {
 	// issues and update the log levels accordingly.
 	for _, logLevelPair := range strings.Split(debugLevel, ",") {
 		if !strings.Contains(logLevelPair, "=") {
-			str := "the specified debug level contains an invalid " +
+			const str = "the specified debug level contains an invalid " +
 				"subsystem/level pair [%v]"
 			return fmt.Errorf(str, logLevelPair)
 		}
@@ -196,14 +196,14 @@ func parseAndSetDebugLevels(debugLevel string) error {
 
 		// Validate subsystem.
 		if _, exists := subsystemLoggers[subsysID]; !exists {
-			str := "the specified subsystem [%v] is invalid -- " +
+			const str = "the specified subsystem [%v] is invalid -- " +
 				"supported subsytems %v"
 			return fmt.Errorf(str, subsysID, supportedSubsystems())
 		}
 
 		// Validate log level.
 		if !validLogLevel(logLevel) {
-			str := "the specified debug level [%v] is invalid"
+			const str = "the specified debug level [%v] is invalid"
 			return fmt.Errorf(str, logLevel)
 		}
 
@@ -312,6 +312,16 @@ func normalizeAddress(addr, defaultPort string) string {
 	return addr
 }
 
+// suppressUsageError signifies that an error that happened during the initial
+// configuration phase should suppress the usage output since it was not caused
+// by the user.
+type suppressUsageError string
+
+// Error implements the error interface.
+func (e suppressUsageError) Error() string {
+	return string(e)
+}
+
 // loadConfig initializes and parses the config using a config file and command
 // line options.
 //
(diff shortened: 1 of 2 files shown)


config.go
```diff
@@ -713,7 +673,7 @@ func loadConfig() (*config, []string, error) {
 	if !cfg.UseLEHTTPS && (!fileExists(cfg.GUITLSCert) || !fileExists(cfg.GUITLSKey)) {
 		err := genCertPair(cfg.GUITLSCert, cfg.GUITLSKey)
 		if err != nil {
-			str := "%s: unable to generate dcrpool's TLS cert/key: %v"
+			const str = "%s: unable to generate dcrpool's TLS cert/key: %w"
 			err := fmt.Errorf(str, funcName, err)
 			fmt.Fprintln(os.Stderr, err)
```
jholdstock, line 678: This Print can go, otherwise its printed twice.

dcrpool.go
```diff
@@ -142,8 +145,16 @@ func realMain() error {
 
 	// Load configuration and parse command line. This also initializes
 	// logging and configures it accordingly.
-	cfg, _, err := loadConfig()
+	appName := filepath.Base(os.Args[0])
+	appName = strings.TrimSuffix(appName, filepath.Ext(appName))
+	cfg, _, err := loadConfig(appName)
 	if err != nil {
+		usageMessage := fmt.Sprintf("Use %s -h to show usage", appName)
```
jholdstock, line 152: This can move down to inside the `if`

config.go
```diff
@@ -516,92 +513,76 @@ func loadConfig() (*config, []string, error) {
...
 	}
 
 	// Create the data directory.
 	err = os.MkdirAll(cfg.DataDir, 0700)
 	if err != nil {
-		str := "%s: unable to create data directory (%s): %v"
+		const str = "%s: unable to create data directory (%s): %w"
 		err := fmt.Errorf(str, funcName, cfg.DataDir, err)
-		fmt.Fprintln(os.Stderr, err)
 		return nil, nil, err
```
jholdstock, line 564: This wasnt showing usage previously, now it is.

config.go
```diff
@@ -713,7 +673,7 @@ func loadConfig() (*config, []string, error) {
 	if !cfg.UseLEHTTPS && (!fileExists(cfg.GUITLSCert) || !fileExists(cfg.GUITLSKey)) {
 		err := genCertPair(cfg.GUITLSCert, cfg.GUITLSKey)
 		if err != nil {
-			str := "%s: unable to generate dcrpool's TLS cert/key: %v"
+			const str = "%s: unable to generate dcrpool's TLS cert/key: %w"
 			err := fmt.Errorf(str, funcName, err)
 			fmt.Fprintln(os.Stderr, err)
 			return nil, nil, err
```
jholdstock, line 679: This wasnt showing usage previously, now it is.

config.go
```diff
@@ -722,9 +682,8 @@ func loadConfig() (*config, []string, error) {
 
 	// Load dcrd RPC certificate.
 	if !fileExists(cfg.DcrdRPCCert) {
-		str := "%s: dcrd RPC certificate (%v) not found"
-		err := fmt.Errorf(str, funcName, cfg.DcrdRPCCert)
-		fmt.Fprintln(os.Stderr, err)
+		err := fmt.Errorf("%s: dcrd RPC certificate (%v) not found", funcName,
+			cfg.DcrdRPCCert)
 		return nil, nil, err
```
jholdstock, line 687: This wasnt showing usage previously, now it is.

config.go
```diff
@@ -486,24 +488,21 @@ func loadConfig() (*config, []string, error) {
...
-		if !errors.As(err, &e) {
-			fmt.Fprintf(os.Stderr, "error parsing config file: %v\n", err)
-			fmt.Fprintln(os.Stderr, usageMessage)
-			return nil, nil, err
+	if preCfg.ConfigFile != defaultConfigFile {
+		err := flags.NewIniParser(parser).ParseFile(preCfg.ConfigFile)
+		if err != nil {
+			var e *os.PathError
+			if !errors.As(err, &e) {
+				err = fmt.Errorf("error parsing config file: %w\n", err)
```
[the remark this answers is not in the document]
davecgh: Good call.  Not to mention it's an error and shouldn't have newlines appended in the error.  Removed.

config.go
davecgh: Removed.

dcrpool.go
davecgh: Done.

config.go
davecgh: Good point.  I was mainly just porting over from dcrd and failure to create the default config file there is just a normal `fmt.Fprintf` instead of an error like it is here.  I missed that it would indeed show the usage in that case.

I suspect it probably should actually be like dcrd and just log an error and move on since it isn't required to exist in order to run.

At any rate, I've modified to suppress it for this PR and we can change the aforementioned behavior in a separate PR if desired.

config.go
davecgh: Suppressed.

config.go
davecgh: Suppressed.

jholdstock (approved):

PR #353, opened 2023-09-12
main: Defer database close.

davecgh:
**This requires #352**.

This defers the database close once opened to ensure it is always properly closed on all error paths.

### DIFF
--- a/dcrpool.go
+++ b/dcrpool.go
@@ -162,6 +162,7 @@ func realMain() error {
 			logRotator.Close()
 		}
 	}()
+	defer mpLog.Info("Shutdown complete")
 
 	var db pool.Database
 	if cfg.UsePostgres {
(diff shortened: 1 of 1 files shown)


dcrpool.go
```diff
@@ -226,7 +226,6 @@ func realMain() error {
 		}
 	}
 
-	db.Close()
 	mpLog.Info("dcrpool shut down.")
```
jholdstock, line 229: Should probably defer this shutdown message as well, otherwise any logging which is included in the deferred tasks will  come **after** the shutdown message.

davecgh: Yeah, I was doing a separate PR that reworks the lifecycle of everything.  I can do it as part of this one though.

davecgh: Added a defer for overall shutdown and updated the existing shutdown message to make it clear it's referring to the hub.

jholdstock (approved):

PR #355, opened 2023-09-12
version: Support single override and add git hash.

davecgh:
This reworks the way versions are handled internally to match other Decred projects such as dcrd and dcrwallet.  Namely, it reverse the semantics such that the individual semver components (major, minor, patch, prerelease, and buildmetadata) are parsed from a full string and exported at init time.

Also, since the version is now parsed and verified to be accurate, it updates the pre-release parsing to properly support dots as required by the spec.

Finally, starting with Go version 1.18, it adds the git commit hash to the version string as buildmetadata when no buildmetadata is otherwise specified.

This provides a few main benefits:

- Allows a single linker override to fully specify the version string instead of having separate ones that can only override the prerelease and build metadata portions
- Provides run-time checks to ensure the full version string is valid per the semver spec regardless of whether it was specified directly in the source or provided via the linker
- The exact commit used to build non-release versions will be in the version string by default

Finally, while here, add some comments regarding the release process to help maintainers and add the `-V` short flag for the version to match other projects for consistency.

### DIFF
--- a/dcrpool.go
+++ b/dcrpool.go
@@ -200,10 +200,10 @@ func realMain() error {
 		}()
 	}
 
-	mpLog.Infof("Version: %s", version())
-	mpLog.Infof("Runtime: Go version %s", runtime.Version())
+	mpLog.Infof("%s version %s (Go version %s %s/%s)", appName,
+		Version, runtime.Version(), runtime.GOOS, runtime.GOARCH)
 	mpLog.Infof("Home dir: %s", cfg.HomeDir)
-	mpLog.Infof("Started dcrpool.")
+	mpLog.Infof("Started dcrpool")
 
 	go func() {
 		select {
--- a/version_nobuildinfo.go
+++ b/version_nobuildinfo.go
@@ -1,0 +1,14 @@
+// Copyright (c) 2021-2022 The Decred developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+//go:build !go1.18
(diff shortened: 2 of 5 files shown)


version_nobuildinfo.go
```diff
@@ -0,0 +1,14 @@
...
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+//go:build !go1.18
+// +build !go1.18
+
+package main
+
+// vcsCommitID returns an empty string for all Go versions prior to 1.18 since
+// the information is not availalbe in binaries prior to that version.
```
jholdstock, line 11:
```suggestion
// the information is not available in binaries prior to that version.
```

dcrpool.go
```diff
@@ -200,10 +200,10 @@ func realMain() error {
 		}()
 	}
 
-	mpLog.Infof("Version: %s", version())
-	mpLog.Infof("Runtime: Go version %s", runtime.Version())
+	mpLog.Infof("%s version %s (Go version %s %s/%s)\n", appName,
```
jholdstock, line 203:
```suggestion
	mpLog.Infof("%s version %s (Go version %s %s/%s)", appName,
```

davecgh: Nice. Removed.

version_nobuildinfo.go
davecgh: Fixed.

jholdstock (approved):

PR #356, opened 2023-09-12
main: Slightly rework main hub and gui lifecycle.

davecgh:
**This is rebased on #355**.

This moves the creation of the context into the main code path, splits the creation logic for the hub and gui into independent funcs, and arranges for the main code path to create those instances and run them.

Since the pool struct is no longer necessary, both it and the associated `newPool` func are removed.

This approach is preferred because it provides the main code path with full control over the lifecycle of the context instead of stuffing it into a separate struct which makes it much harder to reason about and has ultimately led to the use of the cancel func deep in the innards of the hub which should not have that level of control since it is not the coordinator and thus does not have enough details to know when it's actually safe or even should be forcing a shutdown.

This commit does not address the latter part since it will require more in-depth changes to decouple it properly.  However, this is a good start at reducing the overly-tight coupling of the lifecycle that exists as a result.

### DIFF
--- a/dcrpool.go
+++ b/dcrpool.go
@@ -27,32 +27,24 @@ import (
 // Conditional compilation is used to also include SIGTERM and SIGHUP on Unix.
 var signals = []os.Signal{os.Interrupt}
 
-// miningPool represents a decred proof-of-Work mining pool.
-type miningPool struct {
-	ctx    context.Context
-	cancel context.CancelFunc
-	hub    *pool.Hub
-	gui    *gui.GUI
-}
-
-// newPool initializes the mining pool.
-func newPool(db pool.Database, cfg *config) (*miningPool, error) {
-	p := new(miningPool)
-	dcrdRPCCfg := &rpcclient.ConnConfig{
+// newHub returns a new pool hub configured with the provided details that is
+// ready to connect to a consensus daemon and wallet in the case of publicly
+// available pools.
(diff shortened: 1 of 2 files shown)


dcrpool.go
```diff
@@ -195,27 +176,57 @@ func realMain() error {
...
+
+		if err := h.FetchWork(ctx); err != nil {
+			return fmt.Errorf("unable to get work from consensus daemon: %w", err)
+		}
+
+		h.Run(ctx)
+		return nil
+	}
+	if err := runHub(ctx, hub); err != nil {
+		// Ensure the GUI is signaled to shutdown.
```
jholdstock, line 225: Despite calling cancel here, the gui has no waitgroup (or any other mechanism of signalling that it has completed shutdown) so there is no guarantee that it will shutdown completely before the process terminates. Thats an existing problem though, and we can fix that in a later PR.

jholdstock (approved):
Given that we expect this code to move again shortly, I've only given this PR a thorough read and not tested it very much. Will test more thoroughly once its closer to its final form.

dcrpool.go
davecgh: Right.  My hope is that separating it out like this helps show the problem with the current design where it passes the cancel func down deep into the innards of something that does not control the overall process and thus shouldn't be able to just pull the rug out from everything as it's doing now.

PR #357, opened 2023-09-12
gui: Move package to internal.

davecgh:
**This is rebased on #356**.

`dcrpool` is an application and thus all of its package not intended for external consumers should not be publicly visible.

To that end, this moves the `gui` package to an `internal` package meaning that it is no longer a part of the exported main module.


jholdstock (changes requested):
There is a line in README which needs to be updated

> Deploying the user interface requires copying the `dcrpool/gui/assets` folder from source to a reachable location

davecgh:
> There is a line in README which needs to be updated
> 
> > Deploying the user interface requires copying the `dcrpool/gui/assets` folder from source to a reachable location

Thanks, updated.  As an aside, this should really be changed over to use `embed` instead imo.

jholdstock (approved):

PR #359, opened 2023-09-14
pool: Correct endpoint waitgroup logic.

davecgh:
**This requires #358**.

This corrects the waitgroup logic in endpoint handling to ensure it properly terminates when the context is canceled.

It also enables the wait at the end of the endpoint test to help prove correctness.

### DIFF
--- a/pool/endpoint.go
+++ b/pool/endpoint.go
@@ -73,7 +73,6 @@ type connection struct {
 type Endpoint struct {
 	listenAddr string
 	connCh     chan *connection
-	discCh     chan struct{}
 	listener   net.Listener
 	cfg        *EndpointConfig
 	clients    map[string]*Client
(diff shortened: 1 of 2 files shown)


jholdstock (changes requested):
There's an issue here which results in dcrpool never shutting down.

1. `connect` is in the process of creating a new client
1. ctx cancelled
1. `disconnect` calls cancel on every client in the list
1. `connect` adds new client to the list, it is never canceled.

I triggered this condition by adding a `time.Sleep` immediately after the call to `NewClient` and CTRL+C'ing dcrpool whilst the sleep was running.

davecgh:
> There's an issue here which results in dcrpool never shutting down.
>     1. `connect` is in the process of creating a new client
>     2. ctx cancelled
>     3. `disconnect` calls cancel on every client in the list
>     4. `connect` adds new client to the list, it is never canceled.
> 
> I triggered this condition by adding a `time.Sleep` immediately after the call to `NewClient` and CTRL+C'ing dcrpool whilst the sleep was running.

I suspect it's something else, because the new client is [created with the context that is already canceled](https://github.com/decred/dcrpool/blob/a35152f15e0ae671644ae5c35e910f3c3d0f58a9/pool/client.go#L163), so it's already canceled.

davecgh:
For reference later when anyone else is reviewing these PRs.  We tracked the issue down offline and @jholdstock submitted #367 to resolve the actual root cause of what he was seeing.

jholdstock (approved):

jholdstock:
Retested with the changes from #367 and this is working well now.

PR #360, opened 2023-09-14
pool: Use bitlen for calculating iterations.

davecgh:
This switches the calculation for the number of iterations to use the pow limit bit length as opposed to a less accurate round trip through a log2 float64 approximation.

### DIFF
--- a/dcrpool.go
+++ b/dcrpool.go
@@ -9,7 +9,6 @@ import (
 	"errors"
 	"fmt"
 	"math"
-	"math/big"
 	"net/http"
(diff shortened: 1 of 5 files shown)


jholdstock (commented):
There are also instances of this which can be updated in `difficulty.go` and `client_test.go`.

davecgh:
> There are also instances of this which can be updated in `difficulty.go` and `client_test.go`.

Updated.

jholdstock (approved):

PR #364, opened 2023-09-14
pool: Gen extra nonce before initialize.

davecgh:
This modifies the new client logic to create the extra nonce before to creating the client and using it during initialization.  Constructors should generally handle everything that can error prior to creating the object since it is wasteful to create an object only to immediately throw it away when there is an error.

### DIFF
--- a/pool/client.go
+++ b/pool/client.go
@@ -150,39 +150,30 @@ type Client struct {
 	wg          sync.WaitGroup
 }
 
-// generateExtraNonce1 generates a random 4-byte extraNonce1
-// for the client.
-func (c *Client) generateExtraNonce1() error {
-	id := make([]byte, 4)
-	_, err := rand.Read(id)
-	if err != nil {
-		return err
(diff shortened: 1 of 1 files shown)


pool/client.go
```diff
@@ -146,39 +146,38 @@ type Client struct {
 	wg          sync.WaitGroup
 }
 
-// generateExtraNonce1 generates a random 4-byte extraNonce1
-// for the client.
-func (c *Client) generateExtraNonce1() error {
+// generateExtraNonce1 generates a random 4-byte extraNonce1 for the client.
+func generateExtraNonce1() (string, error) {
```
jholdstock, line 150: I don't think this needs to be a func, it isn't reused anywhere. The logic can just be moved into `NewClient` and explained with a comment.

davecgh: Good idea.  Done.

jholdstock (approved):

PR #368, opened 2023-09-14
harness: Make the harness more Windows friendly.

davecgh:
Msys automatically transforms paths for executed commands, but it is unable to do that for paths in the config files.  This updates all of the paths in the config files to use relative paths instead so they work with the various processes.

It also use the posix compliant shebang for locating bash, adds a new variable to control whether or not the profiling ports are enabled, and disables postgres by default.

### DIFF
--- a/harness.sh
+++ b/harness.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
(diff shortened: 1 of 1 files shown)


davecgh:
Tested on linux as well to ensure it still works properly there.

dajohi (approved):

PR #369, opened 2023-09-14
pool: Require TLS >= 1.2 for wallet connection.

davecgh:


jrick:
TLS 1.2 is the default minimum already, and this would only lower the minimum potentially in newer Go versions.

When building with these newer Go releases, it may make sense to remove this config setting entirely.

davecgh:
That's a fair point.  The opposite is also true where compiling it with an older version of Go would lower it.  I personally prefer to be explicit and it also avoid issues with security linters, but I don't really feel super strongly one way or the other about it.

dajohi (approved):

jrick:
with go 1.17 the current minimum, sure this change makes sense.

but if we were to bump the minimum go to go 1.20.x, i don't see any advantages to specifying the minimum TLS version, and only potential downsides in the future.

no qualms with this PR as is as long as 1.17 remains a potential build target (it defaults to TLS 1.0 as a minimum), but that should probably be looked at separately.  And if that were addressed, that would potentially change my OK on this PR :)

PR #371, opened 2023-09-15
gui: Rework create and run lifecycles.

davecgh:
**This requires #370**.

This consists of a series of commits that ultimately make the gui webserver respect the context and perform an orderly shutdown when it is canceled as well 

* First, it consolidates the logic that creates the GUI instance.

  It's generally good practice to first create everything that can fail and then create the final instance at once with the results versus doing it piecemeal.

  Piecemeal creation is typically more error prone and, while not a huge concern here, it also ends up needlessly creating objects that are just thrown away in the event of a later error.

* Second, it initializes the cache when GUI is created.

  It accomplishes this by updating the init func to accept the config instead of individual components and moving the logic that acquires the details via the config from the `Run` method into it.

  It also unexports the `InitCache` method since it is only ever used internal to the `gui` package.

  Not only does it further consolidate the creation logic in one place, it also exposes any errors back to the main thread when the pool is first initializing so any issues can be corrected immediately.

* Third, it splits out the code that periodically updates the cache and notifies websocket clients into a separate method that is invoked by `Run` and introduces a waitgroup for it.

* Fourth, itm oves the code that starts the webserver listeners into a separate method that is invoked by `Run` and removes the `server` field from the GUI type since it is only needed locally in the new method.

* Finally, it makes the webserver respect the context so it is shutdown in an orderly fashion when the context is canceled and updates the gui `Run` method to add the goroutine to the waitgroup now that it will be shutdown as expected.

### DIFF
--- a/internal/gui/gui.go
+++ b/internal/gui/gui.go
@@ -10,6 +10,7 @@ import (
 	"crypto/tls"
 	"errors"
 	"html/template"
+	"net"
 	"net/http"
 	"os"
 	"path/filepath"
@@ -235,6 +236,22 @@ func NewGUI(cfg *Config) (*GUI, error) {
 //
 // It must be run as a routine.
 func (ui *GUI) runWebServer(ctx context.Context) {
+	// Create base HTTP/S server configuration.
+	server := http.Server{
+		// Use the provided context as the parent context for all requests to
+		// ensure handlers are able to react to both client disconnects as well
+		// as shutdown via the provided context.
+		BaseContext: func(l net.Listener) context.Context {
+			return ctx
+		},
+
+		WriteTimeout: time.Second * 30,
+		ReadTimeout:  time.Second * 30,
+		IdleTimeout:  time.Second * 30,
+		Addr:         ui.cfg.GUIListen,
+		Handler:      ui.router,
+	}
+
 	switch {
 	case ui.cfg.UseLEHTTPS:
 		certMgr := &autocert.Manager{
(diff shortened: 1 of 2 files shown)


jholdstock:
Haven't given this a full review yet, but just want to give you a heads up that I still don't think this will be the final form of the web server lifecycle. There is an existing issue which I think should be addressed.

In the case where the GUI is configured to listen on a port which is already in use (a reasonably common scenario) `runWebServer` will return immediately. The rest of the dcrpool process will continue to run. IMO the process should exit(1) in this scenario so the admin can investigate and fix it.

jholdstock:
In [this prospective vspd commit](https://github.com/decred/vspd/commit/4d3e5727d1b79280c8d66ffe9d781e5be2bede46) I am checking that the port is available by starting the TCP listener in `.New`, then starting the server itself in `.Run`. Any thoughts on that approach?

internal/gui/gui.go
```diff
@@ -214,212 +193,207 @@ func (ui *GUI) loadTemplates() error {
...
 	go func() {
-		switch {
-		case ui.cfg.UseLEHTTPS:
-			certMgr := &autocert.Manager{
-				Prompt:     autocert.AcceptTOS,
-				Cache:      autocert.DirCache("certs"),
-				HostPolicy: autocert.HostWhitelist(ui.cfg.Domain),
-			}
+		<-ctx.Done()
+		server.Shutdown(ctx)
```
jholdstock, line 257: Afaict from the doco, calling Shutdown with an already canceled context is about as good as not calling it at all. It needs to be given a new context, probably a `context.WithTimeout`, so it has time to do its work. A few seconds is probably adequate for any in-flight requests to complete.

davecgh: I'm fairly certain it's correct and I've also tested that it does work properly.

The call to `Shutdown` is really only necessary to close the listener(s) so the server returns.  In reality, we wouldn't really need to pass the context at all, but it is a required argument.

From the docs, emphasis mine:

> Shutdown gracefully shuts down the server **without interrupting any active connections**. Shutdown works by first **closing all open listeners**, then closing all **idle** connections, and then **waiting indefinitely for connections to return to idle** and then shut down.

> If the provided context expires before the shutdown is complete, Shutdown returns the context's error, otherwise it returns any  error returned from closing the Server's underlying Listener(s).

So, it waits for in-flight requests to complete regardless of context cancellation.  That is why there is also a `BaseContext` which the handlers themselves can use to exit early.

As an aside, I know this to be true from dcrd's RPC server some time ago.  Before many other change and prior to the addition of the `BaseContext`, if you were to kick off a rescan (or any other request that took a long time to complete) and then shutdown dcrd, the http server would hang until the rescan completed because the rescan handler wasn't checking the context and exiting early.

davecgh:
> Haven't given this a full review yet, but just want to give you a heads up that I still don't think this will be the final form of the web server lifecycle. There is an existing issue which I think should be addressed.
> 
> In the case where the GUI is configured to listen on a port which is already in use (a reasonably common scenario) `runWebServer` will return immediately. The rest of the dcrpool process will continue to run. IMO the process should exit(1) in this scenario so the admin can investigate and fix it.

Right.  I intentionally maintained the existing behavior for that since I was trying to avoid changing too many semantics at once.

That said, I do agree with you.  Although the pool can technically work without the GUI, it wouldn't realistically be very useful without it.  I very much prefer the fail early approach so that sysadmins can resolve any issues immediately versus having to notice an error in a log message and then wondering why the GUI doesn't seem to be working.

davecgh:
> In [this vspd PR](https://github.com/decred/vspd/pull/434) I am checking that the port is available by starting the TCP listener in `.New`, then starting the server itself in `.Run`. Any thoughts on that approach?

Yes, that is almost precisely what I do in dcrd:

```Go
func newServer(ctx context.Context, listenAddrs []string, db database.DB, ....) (*server, error) {
...
	if !cfg.DisableListen {
		var err error
		listeners, nat, err = initListeners(ctx, chainParams, amgr, listenAddrs,
			services)
		if err != nil {
			return nil, err
		}
		if len(listeners) == 0 {
			return nil, errors.New("no valid listen address")
		}
	}
```

internal/gui/gui.go
jholdstock: I've been [playing around](https://go.dev/play/p/-F2Mrj8ft2Z) with this and you're totally right. The server shuts down cleanly regardless of the status of the passed context. This PR stays as it is, I will update decred/vspd#434.

jholdstock: You still might want to log any errors returned by Shutdown?

davecgh: > You still might want to log any errors returned by Shutdown?

I'm not super opposed to it if you feel strongly about, but I generally tend to avoid logging errors the admin can't do anything about anyway.  In my experience, doing that ends up in support tickets.

davecgh: > I've been [playing around](https://go.dev/play/p/-F2Mrj8ft2Z) with this and you're totally right. The server shuts down cleanly regardless of the status of the passed context. This PR stays as it is, I will update [decred/vspd#434](https://github.com/decred/vspd/pull/434).

Here is slightly modified version that shows why the base context is needed too:

https://go.dev/play/p/FbWkwO9DjrR

jholdstock:
All my questions above are answered, and the code is looking good apart from one last thing.`server.Shutdown` doco says:

> Make sure the program doesn't exit and waits instead for Shutdown to return.

An easy way to do that is to swap the code which is blocking and the code in the goroutine. So that would something look like...

```
go server.Listen()
<-ctx.Done()
server.Shutdown()
return
```

davecgh:
> All my questions above are answered, and the code is looking good apart from one last thing.`server.Shutdown` doco says:
> 
> > Make sure the program doesn't exit and waits instead for Shutdown to return.

Updated.  I still think all of the listener logic should be in the constructor or even in the main method and passed in, but that can be another PR.

jholdstock (approved):

PR #373, opened 2023-09-15
pool: Consolidate waitgroup logic.

davecgh:
This switches client, hub, endpoint over to use the new pattern that consolidates the waitgroup logic in a single location.

This pattern is easier to reason about and less error prone since it's trivial to see at a glance that the calls to `Done` are happening as intended versus having to chase them down all over the code.


jholdstock (approved):
Massive improvement.

PR #375, opened 2023-09-15
pool: Prevent potential read goroutine leak.

davecgh:
**This requires #373**.

This ensures the send to the read channel for a client respects the context since it could otherwise end up hanging forever when sending to the channel that no longer has a goroutine reading from it thereby leaking the goroutine.

It also includes the read goroutine in the overall waitgroup for the client to help ensure it shuts down as intended.

### DIFF
--- a/pool/client.go
+++ b/pool/client.go
@@ -741,7 +741,12 @@ func (c *Client) read() {
 			c.cancel()
 			return
 		}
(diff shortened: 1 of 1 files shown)


jholdstock (approved):
Looks good, needs rebase.

PR #376, opened 2023-09-15
pool: Cleanup client read error logging.

davecgh:
**This is rebased on #375**.

This switches the code that examines read errors for timeout to use the proper interface from the net package that is designed for that purpose instead of digging into concrete types.

It also ensures clients that timeout without ever identifying themselves have their uniquely-assigned extra nonce used as the id instead so the log messages can be attributed to the ip of the offending clients.

Finally, it no longer logs an error when an `io.EOF` is encountered as that is an expected condition when the client disconnects and is not an actual error.

### DIFF
--- a/pool/client.go
+++ b/pool/client.go
@@ -183,6 +183,11 @@ func (c *Client) shutdown() {
 	id := c.id
 	c.mtx.RUnlock()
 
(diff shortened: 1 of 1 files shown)


jholdstock (approved):
Good after rebase.

PR #377, opened 2023-09-16
gui: Create listener in GUI constructor.

davecgh:
This creates the listener in the GUI constructor so any issues with listening will cause the entire pool to fail to start thereby allowing the admin to resolve any issues immediately.


jholdstock (approved):

jholdstock:
Referring back to #371, I've confirmed that with this change the dcrpool process will now terminate if the GUI listen port is already in use

PR #378, opened 2023-09-17
pool: Don't give chainstate entire process cancel.

davecgh:
The pool is currently designed to shutdown the entire process deep down in the chainstate when some errors it deems to be fatal occur.  It accomplishes this by passed the cancel func for the entire process all the way down to the chainstate.  Unfortunately, this is not a good design because goroutines deep in the guts of subsystems should not have the ability to directly pull the rug out from under the under process without the upper layers having a chance to do anything about it.

Moreover, some of those errors could actually be temporary errors such as temporary loss of connection to the database or dcrd which then result in the entire process shutting down.

This is a first pass at improving the situation slightly by no longer passing the entire process's cancel func down the chainstate handler and instead arranges for the chainstate handler to return an error back to the hub's Run method which itself creates its own child context that is then canceled to cause the hub to shutdown.

It also corrects a few other issues along the way such as:

- some errors did not have the correct parameters
- the notification callback handlers were not respecting the context which could lead to goroutine leaks
- receivers should never close channels, only senders once they are sure there are no more goroutines that can write to the channel

For now, the same process shutdown behavior is maintained by having the main process shut itself down upon observing the hub shutdown which ensures all other subsystems shutdown properly too.

In other words, the overall behavior is the same, aside from the aforementioned fixes, but this change makes it feasible to further improve the behavior in the future so the pool won't just die due to what are likely temporary failures.

### DIFF
--- a/dcrpool.go
+++ b/dcrpool.go
@@ -30,7 +30,7 @@ var signals = []os.Signal{os.Interrupt}
 // newHub returns a new pool hub configured with the provided details that is
 // ready to connect to a consensus daemon and wallet in the case of publicly
 // available pools.
-func newHub(cfg *config, db pool.Database, cancel context.CancelFunc) (*pool.Hub, error) {
+func newHub(cfg *config, db pool.Database) (*pool.Hub, error) {
 	dcrdRPCCfg := rpcclient.ConnConfig{
 		Host:         cfg.DcrdRPCHost,
 		Endpoint:     "ws",
@@ -65,7 +65,7 @@ func newHub(cfg *config, db pool.Database, cancel context.CancelFunc) (*pool.Hub
 		ClientTimeout:         cfg.clientTimeout,
 	}
 
-	return pool.NewHub(cancel, hcfg)
+	return pool.NewHub(hcfg)
 }
 
 // newGUI returns a new GUI configured with the provided details that is ready
@@ -191,7 +191,7 @@ func realMain() error {
 
 	// Create a hub instance and attempt to perform initial connection and work
 	// acquisition.
-	hub, err := newHub(cfg, db, cancel)
+	hub, err := newHub(cfg, db)
 	if err != nil {
 		mpLog.Errorf("unable to initialize hub: %v", err)
 		return err
@@ -216,7 +216,20 @@ func realMain() error {
 	var wg sync.WaitGroup
 	wg.Add(2)
 	go func() {
+		// Ensure the overall process context is cancelled once Run returns
+		// since the pool can't operate without it and it's possible the hub
+		// shut itself down due to an error deep in the chain state.
+		//
+		// The entire pool really generally shouldn't be shutting itself down
+		// due to chainstate issues and should instead go into a temporarily
+		// disabled state where it stops serving miners while it retries with
+		// increasing backoffs while it attempts to resolve whatever caused it
+		// to shutdown to begin with.
+		//
+		// However, since the code is currently structured under the assumption
+		// the entire process exits, this retains that behavior.
 		hub.Run(ctx)
+		cancel()
 		wg.Done()
 	}()
 	go func() {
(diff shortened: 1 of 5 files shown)


dcrpool.go
```diff
@@ -216,7 +216,20 @@ func realMain() error {
...
 	wg.Add(2)
 	go func() {
+		// Ensure the overall process context is cancelled once Run returns
+		// since the pool can't operate without it and it's possible the hub
+		// shut itself down due to an error deep in the chain state.
+		//
+		// The entire pool really generally shouldn't be shutting itself down
+		// due to chainstate issues and should instead go into a temporarily
+		// disabled state where it stops serving miners while it retries with
+		// increasing backoffs while it attempt to resolve whatever caused it to
```
jholdstock, line 226:
```suggestion
		// increasing backoffs while it attempts to resolve whatever caused it to
```

davecgh: Updated.

pool/chainstate_test.go
```diff
@@ -231,7 +231,10 @@ func testChainState(t *testing.T) {
 	var wg sync.WaitGroup
 	wg.Add(1)
 	go func() {
-		cs.handleChainUpdates(ctx)
+		err := cs.handleChainUpdates(ctx)
+		if err != nil {
+			t.Logf("unexpected error from chain update handler: %v", err)
```
jholdstock, line 236: Why not `t.Fatalf` here?

pool/hub.go
```diff
@@ -623,14 +631,24 @@ func (h *Hub) shutdown() {
...
 		wg.Done()
 	}()
 	go func() {
-		h.chainState.handleChainUpdates(ctx)
+		err := h.chainState.handleChainUpdates(ctx)
+		if err != nil {
+			// Ensure the context is canceled so the remaining goroutines exit
+			// when there was an error that caused the chain update handler to
+			// exit prematurely.
+			log.Error(err)
```
jholdstock, line 649: Maybe skip this log if `errors.Is(err, context.Canceled)`?

pool/chainstate_test.go
davecgh: You can't invoke `t.Fatalf` from goroutines other than the main test goroutine.

pool/hub.go
davecgh: `handleChainUpdates` doesn't return `ctx.Err()` since it's expected that the context will be closed.  Since it doesn't return that, it can never be `context.Canceled`.

jholdstock: That is true if `handleChainUpdates` returns via the `case <-ctx.Done():` path, but it might return via another path.

E.g. if the ctx is canceled during the call to `cs.cfg.GetBlock(ctx, &header.PrevBlock)`, that func might return `context.Canceled`, which in turn will get returned out of `handleChainUpdates`.

pool/chainstate_test.go
jholdstock: Hm. So an error returned by `handleChainUpdates` will cause the test to time out (which results in a panic). We don't specify a timeout so the default of 10 minutes will be used. I guess that's fine because it will still get caught by `run_tests.sh` and thus CI, but its kind of annoying if you hit it during development. Perhaps an `os.Exit(1)`?  Its a bit heavy handed, but I dont think it matters given that this is test code and we already know the test is failing.

davecgh: I changed it over to send the error to a buffered channel that the test checks at the end and fatals if it's non-nil.  In practice, if it exits early the test will failure due to other reasons, so it isn't strictly necessary, but it's safer to check the error.

davecgh: I updated the tests to respect the context when sending and receiving from the channel so that it will allow fatal test errors immediately upon context cancel.  This approach is more robust against other types of failures as well.

As an aside, tests should almost never be blindly sending to and receiving from internal channels without also selecting across the context.

jholdstock: Nice!

jholdstock (approved):

PR #379, opened 2023-09-17
pool: Remove unneeded client disconnect goroutine.

davecgh:
This removes the disconnect goroutine from the endpoint since it is not needed given each client is created with the context and shuts itself down when the context is canceled.

While here, this also updates each client to invoke their associated cancel function when shutting down to ensure it's always called as recommended.


jholdstock (approved):
Looks good. Confirmed that everything cancels correctly in both the case where the pool shuts down and the case where the client disconnects.

PR #381, opened 2023-09-18
main: Support SIGTERM on Win and all unix variants.

davecgh:
Go 1.14 added runtime support for handling the windows `CTRL_CLOSE_EVENT`, `CTRL_LOGOFF_EVENT`, and `CTRL_SHUTDOWN_EVENT` events by adding a definition for `syscall.SIGTERM` to windows and converting the events to that signal. It also added definitions for other common signals, including `SIGHUP`, and treats them as NOOPs on windows.

Consequently, this modifies the logic that deals with conditionally handling `SIGTERM` and `SIGHUP` to handle them for windows as well as all unix-like operating systems, including newer ones that are supported by Go since the code was originally written.

Although the additional signals could probably just be added unconditionally without too much issue now due to the runtime adding stubs to all operating systems the project officially supports, it is safer to be explicit when dealing with syscall definitions since there is no guarantee that they will exist for newly-added operating systems.

Stated more plainly, this change means dcrpool will now be shutdown cleanly on more variants of unix as well as when being terminated by windows itself due to various things such as the user logging off, the window terminal being closed, and the system shutting down.

It also pulls in the logic from dcrd so that logging is added when a signal is received so it is clear why the process is being shutdown.


dcrpool.go
```diff
@@ -139,8 +139,12 @@ func realMain() error {
 		}
 	}()
 
+	// Get a context whose done channel will be closed when a shutdown signal
+	// has been triggered from an OS signal such as SIGNINT (Ctrl+C) or when the
```
jholdstock, line 142:
```suggestion
	// has been triggered from an OS signal such as SIGINT (Ctrl+C) or when the
```

dcrpool.go
```diff
@@ -139,8 +139,12 @@ func realMain() error {
...
+	// returned cacnel function is manually called.
```
jholdstock, line 143:
```suggestion
	// returned cancel function is manually called.
```

dcrpool.go
```diff
@@ -139,8 +138,12 @@ func realMain() error {
 		}
 	}()
 
+	// Get a context whose done channel will be closed when a shutdown signal
```
jholdstock, line 141: Theres a couple of lines at the start of `realMain`  which can be removed, as well as a global declaration of `signals` just under the package imports.

jholdstock (changes requested):

signal.go
```diff
@@ -0,0 +1,44 @@
...
+		case sig := <-interruptChannel:
+			mpLog.Infof("Received signal (%s).  Shutting down...", sig)
+			cancel()
+		case <-ctx.Done():
+		}
+
+		// Listen for repeated signals and display a message so the user knows
+		// the shutdown is in progress and the process is not hung.
+		for {
+			sig := <-interruptChannel
```
jholdstock, line 38:
```suggestion
			sig = <-interruptChannel
```

dcrpool.go
davecgh: Fixed.

jholdstock (approved):

PR #387, opened 2023-09-20
pool: Make endpoint tests more robust and fix race.

davecgh:
This modifies the endpoint tests to respect the context when sending to and receiving from the endpoint channels similar to what was recently done for the chainstate tests.

It also fixes a race in the test due to the add callback being invoked in the run goroutine.

### DIFF
--- a/pool/endpoint_test.go
+++ b/pool/endpoint_test.go
@@ -28,6 +28,7 @@ func makeConn(listener *net.TCPListener, serverCh chan net.Conn) (net.Conn, net.
 }
 
 func testEndpoint(t *testing.T) {
+	const maxConnsPerHost = 3
 	powLimit := chaincfg.SimNetParams().PowLimit
 	iterations := math.Pow(2, float64(256-powLimit.BitLen()))
 	maxGenTime := time.Second * 20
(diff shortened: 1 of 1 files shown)


pool/endpoint_test.go
```diff
@@ -91,7 +91,18 @@ func testEndpoint(t *testing.T) {
 		endpoint.run(ctx)
 		wg.Done()
 	}()
-	time.Sleep(time.Millisecond * 100)
+	sendToConnChanOrFatal := func(msg *connection) {
+		select {
+		case endpoint.connCh <- msg:
+		case <-ctx.Done():
+			t.Fatalf("unexpected enpoing shutdown")
```
jholdstock, line 98:
```suggestion
			t.Fatalf("unexpected endpoint shutdown")
```

pool/endpoint_test.go
```diff
@@ -91,7 +91,18 @@ func testEndpoint(t *testing.T) {
...
+		}
+		select {
+		case <-msg.Done:
+		case <-ctx.Done():
+			t.Fatalf("unexpected enpoing shutdown")
```
jholdstock, line 103:
```suggestion
			t.Fatalf("unexpected endpoint shutdown")
```

pool/endpoint_test.go
davecgh: Fixed.

pool/endpoint_test.go
davecgh: Fixed.

jholdstock (approved):

PR #388, opened 2023-09-20
pool: Break on failed pool tests.

davecgh:
Any calls to `t.Fatal` inside of a `t.Run` will only fatal that specific instance of run.  This adds logic to stop running the remaining pool tests when there is a failed test.

### DIFF
--- a/pool/pool_test.go
+++ b/pool/pool_test.go
@@ -111,6 +111,10 @@ func TestPool(t *testing.T) {
 
 	// Run all tests with bolt DB.
 	for testName, test := range tests {
+		if t.Failed() {
+			break
+		}
+
 		boltDB, err := setupBoltDB()
 		if err != nil {
 			t.Fatalf("setupBoltDB error: %v", err)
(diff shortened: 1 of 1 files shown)


jholdstock:
There is already a built-in flag `-failfast` which you can pass to `go test` for this behaviour. Hard-coding it  and removing the togglability seems a bit heavy handed. Personally I like to know how many tests are broken so I can review them all in one pass, rather than having to repeatedly "fix one test, run tests, fix one test, run tests, etc"

```no-highlight
-failfast
    Do not start new tests after the first test failure.
```
from https://pkg.go.dev/cmd/go/internal/test

davecgh:
I'm aware of the flag, but I'm not sure how else we can avoid purging the database on a test failure (which #389 does) without adding these.

In general, I like to see all failing tests too, but the way these tests are written is not at all conducive to that.  It cost me quite a lot of time not being able to see the data left in the database.

I'm open to another method, but after having just suffered through it, I really don't think we want the database purged on test failures.  It makes it exceedingly difficult to analyze test failures when you nuke the data.

davecgh:
Alright, I thought of a way to allow `-failfast` to work and still allow #389 to retain the data but not otherwise.  Closing and will update #389 accordingly.

PR #389, opened 2023-09-20
pool: Leave failed postgres db state in pool tests.

davecgh:
This modifies the pool tests to avoid cleaning up the postgres data when there is a test failure and the `-failfast` test flag is set so the contents of the database can be examined.

Since a failed run with `-failfast` will no longer purge the data, it also adds logic to purge the data prior to running the tests to ensure any previous failed runs are cleaned up on subsequent runs.

### DIFF
--- a/pool/pool_test.go
+++ b/pool/pool_test.go
@@ -5,6 +5,7 @@
 package pool
 
 import (
+	"flag"
 	"os"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


dajohi (approved):

jholdstock (approved):
I definitely think this is better than hard-coding the fast failure, thanks.

PR #392, opened 2023-09-22
pool: Improve payment UUID with random value.

jholdstock:
The UUID used for the payment ID was not created with any randomness to it, leading to occasional test failures due to duplicated values. Appending a random value to the UUID helps to ensure uniqueness.

The commit also updates the database upgrade code to ensure the old payment ID format is used when performing the old version upgrade.

### DIFF
--- a/pool/hub.go
+++ b/pool/hub.go
@@ -6,7 +6,7 @@ package pool
 
 import (
 	"context"
-	"crypto/rand"
+	crand "crypto/rand"
 	"crypto/tls"
 	"crypto/x509"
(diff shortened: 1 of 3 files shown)


davecgh (approved):

jholdstock:
Rebased this on #391 which means all of the changes to `boltupgrades.go` and `boltupgrades_test.go` are now gone. Everything else remains the same.

davecgh (approved):
Still looks good.  I haven't seen any more failures running the tests several times.

PR #394, opened 2023-09-25
pool: Don't potentially leak paymgr goroutine.

davecgh:
This ensures the goroutine in the payment manager that handles sending rescan responses respects the context so it does not potentially leak the goroutine if the context is canceled prior to receiving a response.

It also corrects the channel close logic by moving into the goroutine. As mentioned in previous commits, receivers should never close channels, only senders once they are sure there are no more goroutines that can write to the channel.

### DIFF
--- a/pool/paymentmgr.go
+++ b/pool/paymentmgr.go
@@ -556,10 +556,14 @@ func fetchRescanResponse(ctx context.Context, rescanSource func() (*walletrpc.Re
 	respCh := make(chan *rescanMsg)
 	go func(ch chan *rescanMsg) {
 		resp, err := rescanSource()
-		ch <- &rescanMsg{
+		select {
+		case <-ctx.Done():
+		case ch <- &rescanMsg{
 			resp: resp,
 			err:  err,
+		}:
 		}
+		close(ch)
 	}(respCh)
 
 	select {
@@ -567,7 +571,6 @@ func fetchRescanResponse(ctx context.Context, rescanSource func() (*walletrpc.Re
 		log.Tracef("%s: context cancelled fetching rescan response", funcName)
 		return nil, errs.ContextCancelled
 	case msg := <-respCh:
-		close(respCh)
 		if msg.err != nil {
 			desc := fmt.Sprintf("%s: unable fetch wallet rescan response, %s",
 				funcName, msg.err)


davecgh:
For reference, the following program illustrates why receivers should never close channels.

```Go
package main

import (
	"context"
	"fmt"
	"sync"
	"time"
)

func main() {
	const senderCloses = true // --------------- TODO: Change to false so the receiver closes the channel.

	start := time.Now()
	defer func() {
		fmt.Println("elapsed:", time.Since(start))
	}()

	// Create two independent contexts that finish at different times to clearly
	// illustrate what can happen when the receiver goroutine notices the
	// context is done first while the sender goroutine simultaneously selects
	// the send channel.
	ctx1, cancel1 := context.WithCancel(context.Background())
	ctx2, cancel2 := context.WithCancel(context.Background())
	time.AfterFunc(time.Millisecond*800, cancel1)
	time.AfterFunc(time.Millisecond*200, cancel2)

	// Create sender goroutine that uses ctx1 and receiver that uses ctx2.
	ch := make(chan int)
	var wg sync.WaitGroup
	wg.Add(2)
	go func() {
		defer wg.Done()
		defer func() {
			if senderCloses {
				close(ch)
				fmt.Println("sender closed channel after", time.Since(start))
			}
		}()
		ticker := time.NewTicker(time.Millisecond * 100)
		defer ticker.Stop()
		var counter int
		for {
			select {
			case <-ctx1.Done():
				fmt.Println("ctx1 done (via ticker wait)")
				return
			case <-ticker.C:
				counter++
				select {
				case <-ctx1.Done():
					fmt.Println("ctx1 done (via send wait)")
					return
				case ch <- counter:
					fmt.Printf("sending %d to channel\n", counter)
				}
			}
		}
	}()
	go func() {
		defer wg.Done()
		defer func() {
			if !senderCloses {
				close(ch)
				fmt.Println("receiver closed channel after", time.Since(start))
			}
		}()
		for {
			select {
			case <-ctx2.Done():
				fmt.Println("ctx2 done")
				return
			case i := <-ch:
				fmt.Println("Received", i)
			}
		}
	}()
	wg.Wait()
}
```

JoeGruffins (approved):

jholdstock (approved):

PR #396, opened 2023-09-25
pool: Correct racy paymgr processing logic.

davecgh:
This corrects the mutex logic in the payment manager that keeps track of whatever or not payments are already being processed.

In particular, the code that is being updated previously checked and set the processing flag under two separate locks of the mutex which could easily result in the first check returning false for two concurrent goroutines followed both both setting it to true and executing which would defeat the purpose of the check.

### DIFF
--- a/pool/paymentmgr.go
+++ b/pool/paymentmgr.go
@@ -180,21 +180,6 @@ func NewPaymentMgr(pCfg *PaymentMgrConfig) (*PaymentMgr, error) {
 	return pm, nil
 }
 
-// isProcessing returns whether the payment manager is in the process of
-// paying out dividends.
-func (pm *PaymentMgr) isProcessing() bool {
-	pm.mtx.Lock()
-	defer pm.mtx.Unlock()
-	return pm.processing
-}
-
-// setProcessing sets the processing flag to the provided boolean.
-func (pm *PaymentMgr) setProcessing(status bool) {
-	pm.mtx.Lock()
-	pm.processing = status
-	pm.mtx.Unlock()
-}
-
 // sharePercentages calculates the percentages due each participating account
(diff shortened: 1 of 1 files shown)


pool/paymentmgr.go
```diff
@@ -726,13 +710,18 @@ func (pm *PaymentMgr) payDividends(ctx context.Context, height uint32, treasuryA
 		return nil
 	}
 
-	if pm.isProcessing() {
+	pm.mtx.Lock()
+	if pm.processing {
+		pm.mtx.Unlock()
```
jholdstock, line 715: Looks like the `return` got lost?

jholdstock (changes requested):
Given that `payDividends` is the only func which accesses the Mutex, I think this can be really simplified. Mutex can just be locked once at the start of the func and then unlocked in a defer.

I don't even think its actually possible for `payDividends` to be called concurrently because it only has one caller which is in a single goroutine, but we can explore that in another PR.

pool/paymentmgr.go
davecgh: Yep. Lost it in a rebase it seems.  Added back.

davecgh:
Looking a little more closely, I see that that `payDividends` is the result of receiving a message on the `paymentCh` which itself  is sent via `processPayments`.  Then `processPayments` is used in the chain state config and is invoked as a goroutine `handleChainUpdates` for connected blocks.

Since blocks can arrive at any time, it is possible that `payDividends` is still running.  Changing it to a mutex for the entire function would mean it will end up running it for every connected block instead of intentionally skipping rapid blocks and processing them as a batch as it does with the current semantics.

jholdstock (approved):

PR #403, opened 2023-10-05
pool: Calculate share weights dynamically.

davecgh:
This calculates the share weights in the global share weights map dynamically based on the entires in the map of the miner hash rates instead of hard-coding them to manually calculated values that can easily become outdated when the hash rates are updated.

### DIFF
--- a/pool/share.go
+++ b/pool/share.go
@@ -13,13 +13,39 @@ import (
 )
 
 // ShareWeights represents the associated weights for each known DCR miner.
-// With the share weight of the lowest hash DCR miner (LHM) being 1, the
-// rest were calculated as:
 //
-//	(Hash of Miner X * Weight of LHM)/ Hash of LHM
-var ShareWeights = map[string]*big.Rat{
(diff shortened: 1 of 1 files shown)


pool/share.go
```diff
@@ -13,13 +13,32 @@ import (
...
 //
-//	(Hash of Miner X * Weight of LHM)/ Hash of LHM
-var ShareWeights = map[string]*big.Rat{
-	CPU: new(big.Rat).SetFloat64(1.0), // Reserved for testing.
-}
+// The weights are calculated as:
+//
+//	Hash of Miner X / Hash of LHM
+var ShareWeights = func() map[string]*big.Rat {
+	// Find the lowest hash rate miner.  When there is not
```
jholdstock, line 21: Incomplete comment

jholdstock (changes requested):

pool/share.go
davecgh: Added a short circuit path for the empty case and commented it accordingly.

jholdstock (approved):

PR #406, opened 2023-10-11
pool: Use regular expressions to identify clients.

davecgh:
This modifies that logic that identifies mining clients by their user agent to use regular expressions instead of requiring hard coded version numbers.

It also adds tests to ensure proper matching functionality.

The primary motivation is to allow the pool to continue to work with new patch versions of mining clients without requiring updates the pool software.

However, it also provides the flexibility to adapt to specific client software that might not conform to semver semantics in the future if is desired.


pool/miner_id.go
```diff
@@ -6,60 +6,34 @@ package pool
...
-	set := make(map[int]string, len(miners))
-	for id, entry := range miners {
-		set[id] = entry
-	}
-	sub := &minerIDPair{
-		id:     id,
-		miners: set,
+	// These regular expressions are used to identify the expected mining
+	// clients by the user agents in their mining.subscribe requests.
+	cpuRE = regexp.MustCompile(`cpuminer/1\.0\.0|[1-9]\d*`)
```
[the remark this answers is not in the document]
davecgh: Updated to make it require strict semantic versioning 2.0.0 and added a bunch of tests.

pool/minerid.go
```diff
@@ -0,0 +1,59 @@
...
+
+var (
+	// These regular expressions are used to identify the expected mining
+	// clients by the user agents in their mining.subscribe requests.
+	cpuRE = newUserAgentRE("cpuminer", 1, 0)
+	nhRE  = newUserAgentRE("NiceHash", 1, 0)
+
+	// miningClients maps regular expressions to the supported mining client IDs
+	// for all user agents that match the regular expression.
+	miningClients = map[*regexp.Regexp][]string{
```
jholdstock, line 42:
```suggestion
	miningClients = map[*regexp.Regexp]string{
```

I think this change would serve to simplify things quite a bit. Maybe there is a reason for it, but I don't see why a single ID would need to be mapping to multiple miner types. We can add support for that later if it ever becomes necessary.

davecgh: I left it that way because there is infrastructure for upgrading based on the hash rate.  There is nothing forcing two different devices to use a different user agent, and as a result, you can have multiple devices with wildly different hashrates identifying themselves the same.

There are tests which cover the upgrading as well.

davecgh: As an aside, I think that ideally the pool really generally shouldn't even be assigning hash rates and instead should figure it out dynamically (as the client upgrade code is doing for devices with the same UA).

Then, it could assign some initial hash rates per UA and even better would be for the protocol to allow clients to specify it themselves (which is unfortunately not part of stratum).

However, that is a fair amount of work that doesn't seem worth it given the goal at the moment is primarily getting into shape for BLAKE3 and gominer as opposed to adding new features.

jholdstock (approved):

PR #408, opened 2023-10-11
multi: Remove no longer relevant blake256 pad code.

davecgh:
This removes the blake256 padding code since it no longer applies.


jholdstock (commented):
One more comment to update, otherwise looks good.

https://github.com/decred/dcrpool/blob/73a5cd59ee348aa0175c5f9fe64bc3c5a167279a/pool/hub.go#L46-L56

jholdstock (approved):

PR #413, opened 2023-10-13
pool: Add gominer 2.0.x support.

davecgh:
**This requires #406**.

This adds gominer version 2.0.x as a supported mining client.

### DIFF
--- a/internal/gui/assets/templates/index.html
+++ b/internal/gui/assets/templates/index.html
@@ -62,8 +62,10 @@
                             </p>
                             <div class="d-flex flex-row">
                                 <ul>
-                                    <li>
-                                        Coming Soon!
+                                    <li><a
+                                        rel="noopener noreferrer"
+                                        target="_blank"
+                                        href="https://github.com/decred/gominer">Gominer&nbsp;2.0.x</a>
                                     </li>
                                 </ul>
                             </div>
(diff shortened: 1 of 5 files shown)


internal/gui/assets/templates/index.html
```diff
@@ -62,8 +62,10 @@ <h2>Identify the Miner</h2>
                             </p>
                             <div class="d-flex flex-row">
                                 <ul>
-                                    <li>
-                                        Coming Soon!
+                                    <li><a
+                                        rel="noreferrer"
```
jholdstock, line 66: We typically use `rel="noopener noreferrer"` on external links. `noopener` prevents the page in the newly opened tab from being able to access/manipulate the originating tab, e.g. it could redirect to a phishing page.

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noopener

davecgh: I can add it if you want, but target="_blank" implicitly adds it per the docs.

davecgh: Added it even though, as mentioned, it was already implicit because of the target.

jholdstock: Nice, I wasn't aware. Still nice to have just for consistency with other links/projects.

jholdstock (approved):

PR #415, opened 2023-10-14
multi: Rework client user agent id logic.

davecgh:
This reworks the client user agent identification logic to improve its efficiency and flexibility as well as to easily support old minor versions.

It does this by changing the matching logic to first parse the user agent into its individual components and then attempting to match against that parsed information using matching functions as opposed to encoding the more specific matching logic directly into a regular expression.

The user agent parsing first attempts to split it into a client name and version part and when that is successful further attempts to parse the version part into the individual semantic version components using a regular expression with capture groups.

The matching functions are closures that accept the parsed user agent details and may impose arbitrary criteria.

For convenience a default matching function is added that requires the user agent to have a provided client name and major version as well as a minor version that is less than or equal to specified value.

The user agent matching tests are updated accordingly.

Finally, `decred-gominer` is updated to support up to version 2.1.x so the pool will work with both version 2.0.0 as well as the master branch that will be moving to version 2.1.0-pre for ongoing development.

### DIFF
--- a/internal/semver/semver_test.go
+++ b/internal/semver/semver_test.go
@@ -1,0 +1,359 @@
+// Copyright (c) 2023 The Decred developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package semver
+
+import "testing"
+
+// TestParse ensures the semantic versioning 2.0.0 parsing logic works as
+// intended.
+func TestParse(t *testing.T) {
+	tests := []struct {
+		version           string // version string to parse
+		wantErr           bool   // whether or not an error is expected
+		wantMajor         uint32 // expected major version
+		wantMinor         uint32 // expected minor version
+		wantPatch         uint32 // expected patch version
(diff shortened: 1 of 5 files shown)


internal/semver/semver_test.go
```diff
@@ -0,0 +1,358 @@
...
+	}, {
+		version: "9.8.7+meta+meta",
+		wantErr: true,
+	}, {
+		version: "9.8.7-whatever+meta+meta",
+		wantErr: true,
+	}}
+
+	for _, test := range tests {
+		// Parse the test version string and ensure the
```
jholdstock, line 321: truncated comment

jholdstock (commented):
Looks good other than comment typo

internal/semver/semver_test.go
davecgh: Updated.

jholdstock (approved):

PR #416, opened 2023-10-18
Use blockHash for AcceptedWorkID, not powHash.

jholdstock:
blockHash is used to construct the AcceptedWorkID when retrieving AcceptedWork from the database, not the powHash.

Because of this bug, dcrpool was failing to register when a block mined by the pool had been confirmed, thus it was never sending the coinbase from that block to the mining participants.

### DIFF
--- a/pool/client.go
+++ b/pool/client.go
@@ -612,9 +612,9 @@ func (c *Client) handleSubmitWorkRequest(ctx context.Context, req *Request, allo
 		return errs.PoolError(errs.WorkRejected, desc)
 	}
 
-	// Create accepted work if the work submission is accepted
-	// by the mining node.
-	work := NewAcceptedWork(powHash.String(), header.PrevBlock.String(),
+	// Create accepted work if the work submission is accepted by the mining
+	// node.
+	work := NewAcceptedWork(header.BlockHash().String(), header.PrevBlock.String(),
 		header.Height, c.FetchAccountID(), miner)
 	err = c.cfg.db.persistAcceptedWork(work)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pool/client.go
```diff
@@ -631,7 +631,7 @@ func (c *Client) handleSubmitWorkRequest(ctx context.Context, req *Request, allo
 		c.sendMessage(resp)
 		return err
 	}
-	log.Tracef("Work %s accepted by the network", hash)
+	log.Tracef("Work %s accepted by the network", header.BlockHash())
```
davecgh, line 634: Calculating block hashes isn't all that expensive, but it seems wasteful to calculate them back to back like this.

Maybe?

```Go
blockHash := header.BlockHash()
work := NewAcceptedWork(blockHash.String(), ...)
...
log.Tracef("Work %s accepted by the network", blockHash)
```

davecgh:
As an aside, that function is rather poorly named and should probably be renamed in a separate PR.  It didn't matter before since the block hash and proof of work hash were the same, but that is no longer the case.  Probably something like `NewAcceptedBlock` instead.

pool/client.go
jholdstock: Multiple other log lines are writing the powHash instead of the block hash, so to maintain consistency I've reverted the logging change.

davecgh:
Good update with the rename to `powHash`.  I was actually thinking the same thing!

pool/client.go
davecgh: Alright.  I'll approve since it's more important to fix the issue for now.

That said, I think there is a distinct difference though that the code is not really properly making in here that will make it very difficult to track what is going on in the future from the logs.

Namely, shares submitted by workers should be based on the proof of work hash.  Meanwhile, _blocks_ submitted to the network should be based on the block hash.

The reason for that is that there is no way to look up a block by its proof of work hash, because it is not intended to be an identifier (that is what the block hash is for).

Given that, I think there really should be a logged mapping of the block hash <-> proof of work hash for blocks submitted to the network.  Otherwise, as mentioned, admins won't be able to easily correlate and track down issues.

davecgh (approved):

PR #422, opened 2023-11-06
pool: Wrap ChainState notifications handlers.

jholdstock:
Wrapping the notification handlers into functions means that the msg.Done channel only needs to be closed in a single place, rather than having it closed before every continue/return (which is annoying and very error prone).

Looks like a huge diff but its mostly indentation.


davecgh (approved):
Much nicer.  This is how I tend to do handlers from goroutines in general since it is less error prone and also requires less indentation.

As an aside, in response to the whitespace comment, github has a config option to hide the whitespace when reviewing a diff like this to make it much easier to review for correctness.

PR #436, opened 2024-08-02
docs: Update for required Go ver and 1.3.0 release.

davecgh:


jholdstock (approved):

jholdstock:
Oops, should have waited until a 1.3 branch was actually tagged. Will get on that now.

davecgh:
Yeah, good point.  I guess we're going to follow the same pattern as dcrd now, so that probably should be `go install github.com/decred/dcrpool@release-v1.3.0` too.

jholdstock:
In fact, 1.3.0 is not correct anyway. The version following 1.2.0 should be 2.0.0 because of the database changes in #391.

PR #437, opened 2024-08-05
Prepare for 2.0.0 release.

jholdstock:


README.md
```diff
@@ -65,7 +65,7 @@ mining, a certificate from an authority (`CA`) like
   command without cloning this repository:
 
   ```sh
-  $ go install github.com/decred/dcrpool@v1.3.0
+  $ go install github.com/decred/dcrpool@v2.0.0
```
davecgh, line 68: This won't work.  The reason we use `release-v2.0.0` is because if you use `v2.0.0` you will have to bump the major module version to include `/v2` in the `go.mod`, meaning this would have to be `go install github.com/decred/dcrpool/v2@v2.0.0`.

davecgh: Unfortunately, I see that the `pool` package is not internal and thus if its API changes in a breaking way, the main `dcrpool` module needs to have its major bumped.

I compared against `v.1.2.0` via `git diff v1.2.0..HEAD pool` and I see that at the very least `pool.ChainStateConfig`, `pool.ClientConfig`, and `pool.EndpointConfig` were changed in a breaking way, so that technically forces the main `dcrpool` module to become `/v2`.

We can probably get away with not bumping it though because there are no external importers of the module.

For what it's worth, this is why there should not be any code exported from the main module at all when the main module is the application itself.  It's a mistake that `dcrwallet` makes too.  It constantly forces major module version bumps which is super annoying for consumers and constantly break install scripts, version linker flags (because they have to become `.../v2.Release` , etc.

The `pool` package should probably be moved to internal to avoid this issue in the future.

davecgh:
Are you planning on creating the `release-v2.0` branch and commit that remove the `-pre` after this?

jholdstock:
Exactly. Following this PR:

- Create a `release-v2.0` branch with `-pre` removed from version
- Tag `release-v2.0.0` from that branch
- Create a PR on master to bump version to `2.1.0-pre`

davecgh (approved):

PR #439, opened 2024-08-07
Unexport everything in the main package(s)

jholdstock:
dcrpool is a standalone application and none of the code in its main package is designed to be reused or imported by other projects.

### DIFF
--- a/version.go
+++ b/version.go
@@ -1,5 +1,5 @@
 // Copyright (c) 2013-2014 The btcsuite developers
-// Copyright (c) 2015-2023 The Decred developers
+// Copyright (c) 2015-2024 The Decred developers
 // Use of this source code is governed by an ISC
 // license that can be found in the LICENSE file.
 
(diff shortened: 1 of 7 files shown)


version.go
```diff
@@ -18,60 +18,60 @@ var (
...
-	// '-ldflags "-X main.Version=fullsemver"'
+	// '-ldflags "-X main.version=fullsemver"'
 	// if needed.
 	//
 	// It MUST be a full semantic version per the semantic versioning spec or
 	// the app will panic at runtime.  Of particular note is the pre-release
 	// and build metadata portions MUST only contain characters from
 	// semanticAlphabet.
-	Version = "1.3.0-pre"
+	version = "1.3.0-pre"
```
davecgh, line 43: These should actually remain exported, because it allows the reproducible build tooling to override things for release builds.

jholdstock: I tested this before opening the PR and it still works.

```
$ go build -ldflags "-X main.version=9.9.9" . && ./dcrpool --version
dcrpool version 9.9.9+e63d74502 (Go version go1.22.5 linux/amd64)
```

davecgh: Hmm, ok.  I wonder if that changed at some point.  I know it didn't work back when the pattern was first implemented.

davecgh (approved):

PR #445, opened 2025-12-29
build: Update Decred deps.

dajohi:
Test against Go 1.25


jrick (approved):
looks fine

probably want to tag a new release with this.

